PR #5, opened 2022-04-03
JavaScript solution for 226-Invert-Binary-Tree

mitchellsullivan:
Should these try to follow line-by-line the original Python solutions, i.e., not introduce complexities like swap syntax (also available in Python, but not used in original solution)?


neetcode-gh:
It doesn't have to be line-by-line, but similar to the python solution would be nice! Thanks for the solution but someone added a JS solution for this problem already!

PR #8, opened 2022-04-03
feat: added JS 217-Contains-Duplicate solution 

DiegoJunges:
I tried to keep the Class Solution standard and made it as close as possible to the video solution, so that people watching the video can understand the Javascript code. :)


neetcode-gh:
Thanks but someone did a JS solution for this.

PR #15, opened 2022-04-03
Added 435 Non-Overlapping Intervals Java Solution

ProgrammingPete:
This solution is tested and passed leetcode tests. Also I am pretty sure it is the best solution (based off Neetcode videos)


neetcode-gh:
Thanks!

PR #18, opened 2022-04-04
java solution for leetcode 217 contains duplicate

sukrut57:
- Added java solution for leetcode 217 contains duplicate
- -Added java solution for leetcode 242 valid anagram

### DIFF
--- a/java/217-Contains-Duplicate.java
+++ b/java/217-Contains-Duplicate.java
@@ -1,0 +1,10 @@
+class Solution {
+    public boolean containsDuplicate(int[] nums) {
+        HashSet<Integer> hashSet=new HashSet();
+        for(int i=0;i<nums.length;i++){
+            if(hashSet.contains(nums[i])) return true;
+            else hashSet.add(nums[i]);
+        }
+        return false;
+    }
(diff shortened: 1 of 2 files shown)


java/217-Contains-Duplicate.java
```diff
@@ -0,0 +1,10 @@
+class Solution {
+    public boolean containsDuplicate(int[] nums) {
+        HashSet<Integer> hashSet=new HashSet();
+        for(int i=0;i<nums.length;i++){
+            if(hashSet.contains(nums[i])) return true;
```
detronetdip, line 5: Insted of doing this you can use hasSet.add(nus[i]).
code will look like->
```
class Solution {
    public boolean containsDuplicate(int[] nums) {
        HashSet<Integer> set=new HashSet<>();
        for(int i=0;i<nums.length;i++){
            if(!set.add(nums[i])){
                return true;
            }
        }
        return false;
    }
}
```

detronetdip (changes requested):

neetcode-gh:
Thanks but someone did the Java solution for this

PR #22, opened 2022-04-04
Create 347-Top-k-frequent-elements.swift

chandra9302:
Swift solution of Top k frequent elements. Leetcode judge accepted the solution.


chandra9302 (commented):
Replaced generic class name with problem specific.

neetcode-gh:
Thanks!

PR #24, opened 2022-04-04
Create 217-Contains-Duplicate.swift

chandra9302:
Swift version of Contains Duplicate. Accepted by leetcode judge.


chandra9302 (commented):
Replace generic class name with problem specific.

PR #26, opened 2022-04-04
Create 1-Two-Sum.swift

chandra9302:
Swift version of Two Sum. Accepted by Leetcode Judge.


swift/1-Two-Sum.swift
```diff
@@ -0,0 +1,14 @@
+class Solution {
```
boudhayan, line 1: Can we remove the `class Solution` as it will create duplicate classes with the same name which will result in a compilation error?

boudhayan (commented):
.

chandra9302 (commented):
Replace the generic class name with problem specific.

PR #60, opened 2022-04-05
Create 271-Encode-and-Decode-Strings.swift

chandra9302:
Swift version of 271-Encode-and-Decode-Strings, tested locally since it's a premium problem on leetcode and lintcode doesn't support swift language.


mitchellirvin:
Hi @chandra9302 ! Thank you for your contribution. Unfortunately another person opened an earlier PR solving this problem, so I'm closing this as a duplicate!

PR #124, opened 2022-04-10
refactor: updated encode and decode strings

rahulshishodia:
## Changes
Updated solution to align it with video explanation and your python solution.


mitchellirvin:
Thanks @rahulshishodia !

PR #131, opened 2022-04-11
Add 42-Trapping-Rain-Water.js

IbuAR:
Added solution for 42-Trapping-Rain-Water in javascript


imaginate:
Solution already exists.

PR #154, opened 2022-04-14
C# - arrays

joriguzman:
New folder for C# solutions, starting with array problems


ShreyasJejurkar:
Why your code is formatted like JS? In C# we keep parenthesis in the next line! 
Maybe you can keep `.editorconfig` in C# folder, so VSCode or most likely VS will pick up that for formatting settings.

joriguzman:
> Why your code is formatted like JS? In C# we keep parenthesis in the next line! Maybe you can keep `.editorconfig` in C# folder, so VSCode or most likely VS will pick up that for formatting settings.

To be consistent with other languages in the repo. It has the same readability, and cuts out a lot of unnecessary space.
C# convention of starting braces on next line is not a hard and fast rule. What's important is the code is readable and the formatting is consistent.

mitchellirvin:
Thanks @joriguzman ! sorry it's taken so long to review

PR #206, opened 2022-04-23
Add Go solutions

bumblebee211196:
1. Add solution for problem 242
2. Add solution for problem 1


mitchellirvin (changes requested):
Hi @bumblebee211196 ! sorry for the delay, do you mind updating the folder naming convention to be `snake_case`? trying to keep things consistent with the other languages. specifically, `arraysAndHashing` would be `arrays_&_hashing` and `blind75` would be `blind_75`

Ahmad-A0:
This is now a duplicate.

PR #223, opened 2022-04-26
Simply 242 Python solution

ethnh:
In Python, you do not have to iterate to prove equality


albertalexandrov (commented):
To simplify more one can use Counter from collections

mitchellirvin:
Thanks @ethnh ! I added this change in the new `/python` dir, we're going to be moving all the python solutions to that dir to keep the root clean

Issue #189, opened 2022-04-18
Adding multiple solutions


gabr0236:
Hey Neetcode, can i add different solutions to existing problems? If so, should i add the solution to the existing solution file?

For me personally, it really helps to look at different solutions to understand the problem completely.

mitchellirvin:
hi @gabr0236 ! i think for problems that have multiple distinct solutions (even if they're not equally efficient) it would be great to include them all.

maybe we should have a comment block for each solve that outlines (at least) the time/space complexity?

mitchellirvin:
@neetcode-gh i figure it's not worth blocking on this. I'll +1 these types of PRs as I review them. we can rewind if you think we shouldn't?

PR #224, opened 2022-04-26
Added two additional methods - 217 Contains Duplicates

qasimala:
I've added two additional methods on how to solve the problem in JavaScript using Set() and another method of using Map() that makes the code more readable. 

I've also added some comments that explain the code. 

**If you're only accepting one solution per problem, please don't complete this pull request.**


Ahmad-A0:
> If you're only accepting one solution per problem, please don't complete this pull request.
I think I'll link #189 so all the discussion about this can happen in one place.

mitchellirvin:
Thanks @qasimala !

PR #239, opened 2022-05-01
C# solution for Find the Duplicate Number

b4ry:
I added two solutions for the https://leetcode.com/problems/find-the-duplicate-number problem.
Both are the same in the Big O terms.

### DIFF
--- a/cs/FindDuplicateNumber.cs
+++ b/cs/FindDuplicateNumber.cs
@@ -1,0 +1,54 @@
+public class Solution
+{
+    /* Slow and fast pointer solution */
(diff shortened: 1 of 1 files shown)


mitchellirvin (changes requested):
do you mind updating the folder to be csharp to match the existing convention, and updating your file name to match the rest of the repo? should like something like Find-Duplicate-Number.cs

mitchellirvin:
closing as stale! please feel free to reopen once those changes are made, provided the problem doesn't already have a solution for csharp!

PR #240, opened 2022-05-01
Added some ruby solutions 

katiie:
Added a ruby folder and 3 solutions for 3 easy array problems


Ahmad-A0:
Geez, sorry it took so long to review this PR but it looks like someone has already committed the solutions you proposed, even though it looks like your PR came first :/

PR #241, opened 2022-05-02
Add Raku implementation for several problems

uzluisf:
* 1-Two-Sum.raku
* 146-LRU-Cache.raku
* 155-Min-Stack.raku
* 217-Contains-Duplicate.raku
* 238-Product-of-Array-Except-Self.raku
* 242-Valid-Anagrams.raku
* 347-Top-K-Frequent-Elements.raku
* 49-Group-Anagrams.raku

### DIFF
--- a/raku/1-Two-Sum.raku
+++ b/raku/1-Two-Sum.raku
@@ -1,0 +1,13 @@
+use v6.c;
+
+class Solution {
+    method twoSum(Int:D @nums, Int:D $target --> Array[Int]) {
+        my Int %seen-diff;
+
+        for @nums.kv -> $i, $num {
+            my $diff = $target - $num;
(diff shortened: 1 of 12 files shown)


mitchellirvin:
@uzluisf hello! thanks for contributing. do you have an easy way for me to verify that these solutions are correct? I don't see Raku (or Perl) as a supported language in LeetCode. If the language isn't supported in LeetCode, I'd be inclined to think it's not worth including in the repo?

uzluisf:
Hi @mitchellirvin 

Sorry for the late reply, the email notification got lost on my email.

>do you have an easy way for me to verify that these solutions are correct?

An easy way would be to write/run some test suite. 

>If the language isn't supported in LeetCode, I'd be inclined to think it's not worth including in the repo?

Fair enough!

PR #243, opened 2022-05-02
this fixes #242

### LINKED ISSUE #242: Javascript 125-ValidPalindrome.js solution is not implemented correctly
As the title says the solution is not correct. There are two problems.
First it should check if the character is not alphanumeric in the while loops. The current solution checks if it is alphanumeric
Second the isAlphaNumeric method is implemented incorrectly.

Martin-Karsten:
this fixes #242

### DIFF
--- a/javascript/125-ValidPalindrome.js
+++ b/javascript/125-ValidPalindrome.js
@@ -1,30 +1,32 @@
-class isValidPalindrome {
-    constructor(string){
-        this.string=string;
-    }
-    isPalindrome(string){
-        let left =0;
-        let right=string.length-1;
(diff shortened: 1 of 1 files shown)


imaginate:
These problems were resolved by an earlier [pull request](https://github.com/neetcode-gh/leetcode/commit/1adba3f77089f8982d480771fe3abe9b6e19c8fb), and the solution was further cleaned up with this [commit](https://github.com/neetcode-gh/leetcode/commit/9a02b70ac517e294d3c6e0b9af85b65c83e6e8d9). Thanks for opening the issue and pull request!

PR #253, opened 2022-05-10
Update 200-Number-of-Islands.py

callofktulu:
This queue is not used in the recursive dfs function. Probably a remnant from the bfs solution shown on the youtube video.


neetcode-gh:
Thanks!

PR #258, opened 2022-05-13
simple change for better complexity

monkey-patcher:


mitchellirvin:
thanks @monkey-patcher !

PR #261, opened 2022-05-15
using tuple in lieu of list

mojo-py:
tuple makes more sense from a readability standpoint since we know a tuple's constituent elements aren't going to be altered during program execution whereas when I see a list this foreshadows a mutation is coming but in this program we do not need a mutable data structure. thus I suggest a tuple instead of a list here.


mitchellirvin:
I added this change to the `python/` dir, since we're going to be moving the solutions from root to there. Thanks @mojo-py !

PR #262, opened 2022-05-16
refinements

mojo-py:


mitchellirvin:
thanks @mojo-py ! I added this change to the `python/` dir since we're moving all the python solutions out of the root dir

PR #264, opened 2022-05-16

Mohmed-Refaay:
Shorted solution and takes less memory


Ahmad-A0:
The [leetcode question](https://leetcode.com/problems/top-k-frequent-elements/) includes a _"Follow up: Your algorithm's time complexity must **be better than O(n log n)**, where n is the array's size."_, since `Array#sort` is `O(n log n)` the solution doesn't fit the constraint.

PR #269, opened 2022-05-20
Update 242-Valid-Anagrams.py

Woodenman23:
I was wondering why you don't solve the problem like this. Is it because it is less efficient?

I am quite new to programming but my way is easier to follow for me.

Thanks for all you do.


mitchellirvin:
Hi @Woodenman23 ! Thanks for contributing!

This solution definitely works, but it's less efficient (sorting is O(n log n), where n is the length of the string) than the existing solution.

PR #270, opened 2022-05-21
no stack overflow in python

mojo-py:


mitchellirvin:
@mojo-py overflow here is referring to integer overflow, not stack overflow!

PR #271, opened 2022-05-21
no need to initialize k var

mojo-py:


mitchellirvin:
@mojo-py if you don't initialize k, you'll get an error trying to return it for referencing a variable that's not in the scope of the function!

mitchellirvin:
in general, before opening PRs for changes I would recommend taking your code change and making sure it passes submission on LeetCode (this code has a runtime error that would be easy to discover when executing)

PR #295, opened 2022-06-07
Use the append method as much as possible

anushibin007:
Using string concatenation is not advisable when we are already using a `StringBuilder` to construct the encoded string. We can easily use the `append` method to append the "#" symbol.


mitchellirvin:
hi @anushibin007 ! thanks for contributing! it looks like this change has already been merged on main

PR #304, opened 2022-06-13
212-Word-Search-II.py Implement removeWord

Trip1eLift:
In question 212 word search II, removeWord is required to pass the newer test case on Leetcode due to longer lists of words and the same time limit.

### DIFF
--- a/212-Word-Search-II.py
+++ b/212-Word-Search-II.py
@@ -10,7 +10,17 @@ class TrieNode:
                 cur.children[c] = TrieNode()
             cur = cur.children[c]
         cur.isWord = True
-
+    
+    def removeWord(self, node, word):
+        node.isWord = False
+        for i in range(len(word)-1, 0, -1):
+            w = word[:i]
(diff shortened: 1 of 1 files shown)


pdgreg:
I came to check if there were any other pull requests before I made my own.

Here's my implementation if there is any preference:

```python
 def pruneWord(self, word) -> None:
    cur: TrieNode = self
    nodeAndChildKey: list[tuple[TrieNode, str]] = []
    for char in word:
        nodeAndChildKey.append((cur, char))
        cur = cur.children[char]

    for parentNode, childKey in reversed(nodeAndChildKey):
        targetNode = parentNode.children[childKey]
        if len(targetNode.children) == 0:
            del parentNode.children[childKey]
        else:
            return
```

mitchellirvin:
it appears an earlier PR implemented this pruning, closing as a dupe! feel free to reopen if there's unique behavior or value to be added to this solution

PR #321, opened 2022-06-17
update-kotlin-arrays-hashing-1-Two-Sum

folkfrek:
Created a folder for Kotlin solutions and added solution for Two-Sum


mitchellirvin:
hi @folkfrek ! thanks for contributing! 

closing this out as the solutions for the other two files are empty, and two sum had an earlier-dated PR merge with a kotlin solution

PR #330, opened 2022-06-20
folder called "javanewcode"    contains  solutions of the leetcode questions in java code .

bikram8:


SharmaTushar1:
Add a single file at a time to the java folder. Also, there are unnecessary imports in the files which are not needed in leetcode. File naming is also not proper. Name files like 1-Two-Sum. Closed.

PR #335, opened 2022-06-23
Add leading zeros to file name to keep in order

abe-101:
Adding leading zeros to the file names keeps the files in order.
Renaming was wasy thanks to this neet code in python  (pun intended)
```python
import sys,os,re

for file in sys.argv[1:]:
    if __file__ in file: continue
    number, name = file.split('-', 1)
    number = '{:0>4}'.format(number)
    new_name = f'{number}-{name}'
    os.rename(file,new_name)
```


mitchellirvin:
hi @abe-101 ! i'm actually a fan of this, but @neetcode-gh's paths on neetcode.io depend on the current file names, and this would be a breaking change. do you mind opening an issue that we can track separately if we feel like this is worth the lift?

abe-101:
Hey @mitchellirvin Its funny you mentioned the dependancy issue as I was litterly gonna ask just that!
A few days a ago I discovered the new login feature on neetcode.io so i started going through grind75. Iit then downed on me the code solution all probably depend on this repo and my PR that changes the name would brake the site...

PR #342, opened 2022-06-27

DamenH:
I find this to be more intuative, though it does require a work around because Array.prototype.fill([]) seems to fill the bucket array with arrays sharing the same memory address.

### DIFF
--- a/javascript/347-Top-K-Frquent-Elements.js
+++ b/javascript/347-Top-K-Frquent-Elements.js
@@ -6,7 +6,7 @@
 var topKFrequent = function(nums, k) {
     let map = new Map();
     let res = [];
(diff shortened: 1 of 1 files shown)


javascript/347-Top-K-Frquent-Elements.js
```diff
@@ -6,7 +6,7 @@
 var topKFrequent = function(nums, k) {
     let map = new Map();
     let res = [];
-    let bucket = new Array(nums.length+1);
+    let bucket = new Array(nums.length+1).fill(null).map(() => []); // to create unique arrays
```
DamenH, line 9: Could be `new Array(nums.length+1).fill().map(() => []);` or `Array.from({length: nums.length+1}, () => [])`

Ahmad-A0:
Thank you, @DamenH!

PR #343, opened 2022-06-28
Remove recalculated value

lnguyen-cpi:
`row` is unnecessary recalculated. Remove it


mitchellirvin:
looks like this has already been merged, thanks @lnguyen-cpi !

PR #344, opened 2022-06-28
Сoncise and faster 125-Valid-Palindrome

hayorov:
Runtime: 92 ms, faster than 24.49% of Python3 online submissions for Valid Palindrome.

Uses string a-z, 0-9, and O(n) extra memory but looks more concise and straightforward with a single while for 2 pointers.

### DIFF
--- a/125-Valid-Palindrome.py
+++ b/125-Valid-Palindrome.py
@@ -1,19 +1,15 @@
+from string import ascii_lowercase, digits
+
+
 class Solution:
     def isPalindrome(self, s: str) -> bool:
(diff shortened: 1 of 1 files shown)


mitchellirvin:
hi @hayorov ! thank you for contributing!

since this actually effectively slows down the runtime (though it doesn't change Big O) because of the extra loop over s and the `.join`, and this deviates more from the video solution, I'm inclined to say there's not a compelling enough reason to merge.

if you have a solution that is distinct (in terms of runtime, or approach), feel free to add it as a separate function underneath the editorial solution in the file!

PR #354, opened 2022-07-03
Performance improvements for 128-longest-consec

bswole:
Looping through the numSet instead of nums list provides much better performance. Also saw performance boost when passing values in a list to max().

### DIFF
--- a/128-Longest-consecutive-sequence.py
+++ b/128-Longest-consecutive-sequence.py
@@ -3,12 +3,12 @@ class Solution:
         numSet = set(nums)
         longest = 0
         
-        for n in nums:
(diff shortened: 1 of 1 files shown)


mitchellirvin (commented):
hi @bswole ! thanks for contributing! 

i'm down with looping over the set (though it doesn't change Big O), but the change to an array shouldn't provide an effective improvement in performance. did you test this with a local benchmark? the leetcode.com performance evaluation is notoriously unreliable (if you click "submit" 5 times for the same solution you'll see as much as a 50%+ difference)

mitchellirvin:
closing as stale! please feel free to reopen once changes have been addressed

PR #356, opened 2022-07-04
Create 42-Trapping-Rain-Water.swift

seungjun-green:
Added a swift version of solution for 42-Trapping-Rain-Water.swift


mitchellirvin:
thanks @seungjun-green !

PR #360, opened 2022-07-06
Update 125-Valid-Palindrome.java

ALKRIS-55:
Used the array to program the valid palindrome problem and achieve a result which was 38% faster and 7% more efficient than the previous solution

### DIFF
--- a/java/125-Valid-Palindrome.java
+++ b/java/125-Valid-Palindrome.java
@@ -1,3 +1,4 @@
+// Method 1
 public boolean isPalindrome(String s) {
         StringBuilder content = new StringBuilder();
         for(int i = 0; i < s.length(); i++) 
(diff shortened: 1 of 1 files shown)


java/125-Valid-Palindrome.java
```diff
@@ -6,4 +7,30 @@ public boolean isPalindrome(String s) {
...
         return value.equals(content.reverse().toString());
+}
+
+
+//         Method 2
+//         Using Array
+//         38% more fast than the above code
+class Solution {
+    public boolean isPalindrome(String s) {
+        boolean flag = true;
```
mitchellirvin, line 18: declaring this variable up here is unnecessary, we can `return false` on line 30 (there's no reason to keep iterating when we encounter a mismatched character), and `return true` on line 34.

mitchellirvin (commented):
hi @ALKRIS-55 ! what did you use to benchmark the performance improvement? leetcode.com's benchmarks are notoriously inaccurate (just re-running the same code can frequently result in a 50%+ difference in runtime performance).

in general, unless the solution is more efficient in terms of Big O, or takes a distinct approach to solving the problem, I'm not in favor of having multiple instances of the "same" solution (I could be convinced otherwise if this approach was strictly better, but StringBuilders just use arrays under the hood)

mitchellirvin:
in general, unless the solution is more efficient in terms of Big O, or takes a distinct approach to solving the problem, I'm not in favor of having multiple instances of the "same" solution (I could be convinced otherwise if this approach was strictly better, but StringBuilders just use arrays under the hood)

PR #365, opened 2022-07-06
Update 1143, 252, 42

miladra:


mitchellirvin (changes requested):
hi @miladra ! thanks for contributing. 

1. 252 already has a solution, did you mean to extend the existing one? (currently you're creating a new file)
2. 253's [method signature on leetcode.com](https://leetcode.com/problems/meeting-rooms-ii/) is `int[][]`, but you have it listed as `List<Interval>`. am i seeing the wrong signature? did leetcode.com change theirs from `List<Interval>`? (I realize this applies to the existing solution, also. I'll work on getting that fixed separately)

miladra:
Hi @mitchellirvin, Thank you.

As for 252, I intended to extend the existing solution. I made a mistake by creating a new file.
As for 253, the question is locked in Leetcode and I don't have a subscription. My code was written using Lintcode.

mitchellirvin:
@miladra ah, thanks for clarifying! that should be fine since the logic is the same. thanks for updating and contributing!

Issue #368, opened 2022-07-06
Alternate C++ solution #98. Validate Binary Search Tree


resyfer:
A C++ solution to the [98.Validate Binary Search Tree](https://leetcode.com/problems/validate-binary-search-tree/) problem which follows Neetcode's explanation in the video.

KC10201:
I think this might be more in line of the implementation provided.
```
class Solution {
public:
    bool isValidBST(TreeNode* root) {
        using def = function<bool(TreeNode*,long,long)>;
        
        def valid = [&] (auto node, long left, long right) {
            if(!node){
                return true;   
            }
            if(!(node->val < right && node->val > left)){
                return false;
            }
            return (valid(node->left, left, node->val) && 
                    valid(node->right, node->val, right));
        };
        
        return valid(root, long(INT_MIN)-1, long(INT_MAX)+1);
    }
};
```

farhan523:
#369 
`
class Solution
{
    vector<int> res;
    bool inorderTraversal(TreeNode *root)
    {
        if (root == NULL)
            return true;
        bool r1, r2;
        if (root->left != NULL && root->left->val >= root->val)
            return false;
        else
            r1 = inorderTraversal(root->left);
        res.push_back(root->val);
        if (root->right != NULL && root->right->val <= root->val)
            return false;
        else
            r2 = inorderTraversal(root->right);
        return r1 && r2;
    }

    public:
        bool isValidBST(TreeNode *root)
        {
            if (!inorderTraversal(root))
                return false;
            for (int i = 1; i < res.size(); i++)
            {

                if (res[i] <= res[i - 1])
                    return false;
            }
            return true;
        }
};
`

resyfer:
Thank you for the solution, but my issue was created for a possible addition of my solution to the question which followed the solution in the video, and I've added that in the commit following the issue

Ahmad-A0:
I'll mention #424 which is also about this problem.

Ahmad-A0:
I'll close this issue as a parallel PR exists.

PR #369, opened 2022-07-06
ADD: Alternate C++ Solution to No. 98 Validate Binary Search Tree

resyfer:
Issue #368

### DIFF
--- a/cpp/neetcode_150/07_trees/validate_binary_search_tree.cpp
+++ b/cpp/neetcode_150/07_trees/validate_binary_search_tree.cpp
@@ -1,12 +1,3 @@
-/*
-    Given root of binary tree, determine if it's valid (left all < curr, right all > curr)
-
-    Inorder traversal & check if prev >= curr, recursive/iterative solutions
-
-    Time: O(n)
-    Space: O(n)
-*/
-
 /**
  * Definition for a binary tree node.
  * struct TreeNode {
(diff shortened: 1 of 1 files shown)


cpp/neetcode_150/07_trees/validate_binary_search_tree.cpp
```diff
@@ -18,57 +9,53 @@
  *     TreeNode(int x, TreeNode *left, TreeNode *right) : val(x), left(left), right(right) {}
  * };
  */
+
+// Need a long long int alias because of the constraints of TreeNode.val
+#define ll long long int
```
KC10201, line 14: unclear why requires long long instead of just a long

cpp/neetcode_150/07_trees/validate_binary_search_tree.cpp
```diff
@@ -18,57 +9,53 @@
...
-        if (prev != NULL && prev->val >= root->val) {
-            return false;
-        }
-        prev = root;
-        
-        if (!inorder(root->right, prev)) {
+        if(root->val > low && root->val < high) {
+            return dfs(root->left, low, min(high, (ll) root->val))
+                && dfs(root->right, max(low, (ll) root->val), high);
+        } else {
```
KC10201, line 56: Since you've already verified that root->val is greater than low, and smaller than high, there's no need to calculate min(high, root->val), and max(low, root->val)

KC10201, line 56: Since you've already verified that root->val is greater than low, and smaller than high, there's no need to calculate ```min(high, root->val)```, and ```max(low, root->val)```.
Recommend changing it to 
```
 return dfs(root->left, low, root->val)
                && dfs(root->right, root->val, high);
```

KC10201 (changes requested):

cpp/neetcode_150/07_trees/validate_binary_search_tree.cpp
resyfer: I had indeed overlooked the fact that calculating min and max were unnecessary. Thank you for the suggestion.

mitchellirvin:
hi @resyfer ! thanks for contributing! 

I think we should keep the iterative solution that was previously commented out (it's helpful to see multiple ways to solve the same problem, and especially helpful to understand iterative tree traversals as well as recursive). do you mind moving it to a public method titled `isValidBstIterative` or the like? otherwise this looks good to go as soon as merge conflicts are resolved.

mitchellirvin (changes requested):
hi @resyfer ! thanks for contributing! 

I think we should keep the iterative solution that was previously commented out (it's helpful to see multiple ways to solve the same problem, and especially helpful to understand iterative tree traversals as well as recursive). do you mind moving it to a public method titled `isValidBstIterative` or the like (and uncommenting it)? otherwise this looks good to go as soon as merge conflicts are resolved.

Ahmad-A0:
I'll close this as stale due to the changes in directory structure.

PR #371, opened 2022-07-06
Simpler. :)

AlvarMarkhester:


Ahmad-A0:
It's a very nice solution, but it will perform worse on average as creating the `Set` means you have to iterate through the whole of `nums` regardless of if you find duplicates or not.

AlvarMarkhester:
> It's a very nice solution, but it will perform worse on average as creating the `Set` means you have to iterate through the whole of `nums` regardless of if you find duplicates or not. Both solutions are technically `O(N)` though.

Thanks for the feedback :)

mitchellirvin (changes requested):
hi @AlvarMarkhester ! thanks for contributing!

I don't think we should remove the existing solution, as it is likely more intuitive for a new programmer (even though it is more verbose). simpler is not always easier to understand!

you're welcome to add your solution beneath the existing solution, though! it is very elegant

Ahmad-A0:
I'll close this as stale due to the conflicts, but if you choose to resolve them and edit your solution to meet the new guidelines I'll merge this in.

PR #374, opened 2022-07-06
Added Python Code for Number of Connected Components In An Undirected…

ironnicko:
You made a video for this in python but there's no code for it in your website.


neetcode-gh:
Thank you so much!

PR #380, opened 2022-07-07
javascript and typescripts solutions

dipti95:
### Following solutions have been added 
- 261 Graph Valid Tree Javascript Solution
- 200 Number of Islands TypeScript Solution 
- 242 Valid Anagram TypeScript Solution

### DIFF
--- a/javascript/261-Graph-Valid-Tree.js
+++ b/javascript/261-Graph-Valid-Tree.js
@@ -1,0 +1,83 @@
+/**
+ * @param {number} n
+ * @param {number[][]} edges
+ * @return {boolean}
+ */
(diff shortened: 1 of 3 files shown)


mitchellirvin (changes requested):
hi @dipti95 ! thank you for contributing!

it seems like 4 of the problems already have solutions in typescript. if your solutions are unique (according to the README's contributing rules), please update the existing files! otherwise, feel free to remove those files and I'm happy to merge the code that doesn't yet have solutions!

also, in the future, please prefer submitting one PR per solution (it makes it easier for us to review them and get them merged quickly!)

PR #382, opened 2022-07-07
remove redundant check

johan456789:
If target is in dp, it would have return because `(t + nums[i]) == target`.


mitchellirvin (approved):

mitchellirvin:
thanks @johan456789 !

PR #386, opened 2022-07-07
Update 55-Jump-Game.py

UWDavidWu:
directly return a boolean instead of using a ternary operator


mitchellirvin:
thanks @UWDavidWu !

PR #396, opened 2022-07-07
Refactor: Proper formatting and match solution exactly with video for solution 684

abstrekt:


mitchellirvin:
thanks @abstrekt !

PR #403, opened 2022-07-07
Adds a separate, more concise solution

andrewcar:


mitchellirvin (changes requested):
Hi @andrewcar! Thanks for submitting this PR.

In general, it's best to favor readability over conciseness for editorial solutions. While the one-liner is nice, it's not as easy to understand for folks who are new to the problem.

If you'd like to, you're more than welcome to add the one-liner solve as a separate function underneath the existing solution

andrewcar:
Gotcha! Will do.

mitchellirvin:
thanks @andrewcar !

PR #420, opened 2022-07-08
go arrays and hashing solutions

Maximilian5189:


Maximilian5189:
There are some other people contributing go solutions on the same topics. I will close my PR so that it does not too confusing.

PR #426, opened 2022-07-09
Refactor JS - Array & Hash

aakhtar3:


javascript/1-Two-Sum.js
```diff
@@ -1,15 +1,22 @@
...
+            const sumIndex = map.get(complement);
+
+            return [ index, sumIndex ]
         }
+
+        map.set(num, index);
     }
-};
+    
+    return [ -1, -1 ]
```
Ahmad-A0, line 21: I don't think `return [-1, -1]` is needed since _"You may assume that each input would have **exactly one solution**"_ for this problem.

javascript/217-Contains-Duplicate.js
```diff
@@ -1,14 +1,8 @@
...
 var containsDuplicate = function(nums) {
-    const numsSet =  new Set()
-    for(const i of nums){
-        if(numsSet.has(i)){
-            return true
-        } 
-        numsSet.add(i)
-    }
-    return false
+    return (new Set(nums)).size !== nums.length
```
Ahmad-A0, line 7: This solution has already been proposed (#371)

javascript/242-Valid-Anagram.js
```diff
@@ -1,34 +1,52 @@
...
+    addCharFrequency(s, map);
+    return subtractCharFrequency(t, map)
     
-    for (let i = 0; i < s.length; i++) {
-        if (map[s[i]]) {
-            map[s[i]]++;
-        } else {
-            map[s[i]] = 1;
-        }
+};
```
Ahmad-A0, line 30: Adding multiple solutions is being discussed here (#189)

Ahmad-A0 (commented):
Adding multiple solutions is being discussed in #189, It's also difficult to decide which refactors make the code easier to understand, even though I am partial to your use of array methods.

I don't know where @neetcode-gh wants to take this.

aakhtar3:
Closing for another [PR](https://github.com/neetcode-gh/leetcode/pull/445)

Issue #424, opened 2022-07-09
Code solutions differ from video.


Ahmad-A0:
It seems some solutions differ from the video, for `268-Missing-Number` this code is shown in the video and appears as the python solution on neetcode.io:
```py
class Solution:
    def missingNumber(self, nums: List[int]) -> int:
        res = len(nums)
        
        for i in range(len(nums)):
            res += (i - nums[i])
        return res
```

However the Java, C++ and proposed Typescript solution (#422) use slightly different algorithms:

**Java:**
```Java
class Solution {
    public int missingNumber(int[] nums) {
        int sum = 0;
        int total = nums.length * (nums.length+1)/2;
        for(int i = 0; i < nums.length; i++){
            sum += nums[i];
        }
        return total - sum;
    }
}
```

**C++:**
```cpp
class Solution {
public:
    int missingNumber(vector<int>& nums) {
        int n = nums.size();
        int result = n;
        for (int i = 0; i < n; i++) {
            result ^= i ^ nums[i];
        }
        return result;
    }
};
```

**Typescript:**
```ts
function missingNumber(nums: number[]): number {
  let sum: number = 0
  let total: number = (nums.length * (nums.length + 1)) / 2
  for (let i = 0; i < nums.length; i++) {
    sum += nums[i]
  }
  return total - sum
}
```

Ahmad-A0:
I'm not sure how much it matters in this case, but it could definitely be confusing for some of the harder problems.

siphc:
Java and Typescript solutions are algorithmically the same to Python's. They calculate the sum of range [0, n] before iteration, as opposed to during iteration (as with Python's case).

Additionally, https://neetcode.io/ groups `268-Missing-Number` as "Bit Manipulation", and the video does cover the bitwise `xor` approach to the problem, which is what the C++ solution implemented.

Ahmad-A0:
Sorry, I should have chosen a better example. I'll edit my initial comment when I find one, but I wanted to raise the issue of how code is translated between the python solution and other languages to avoid confusion and refactors later on.

PR #432, opened 2022-07-09
Updating 125-valid-Palindrome.java so it uses two pointers

vorenusCoA:
The current solution works fine but it differs with the one explained on the video. Since the exercise is under "Two Pointers" section I think it would be nice if the solution shows that.


Ahmad-A0:
I'll link this to #424.

PR #435, opened 2022-07-09
Solution for 347. Top K Frequent Elements

parulgarg123:
Solution for 

347. Top K Frequent Elements


mitchellirvin:
hi @parulgarg123 ! thank you for contributing!

it looks like your new solution is the same as the existing `Solution1`, and your change deletes a more optimal solution using Hoare's algorithm. for those reasons I don't think we should merge this PR. 

please let me know if I can clarify anything, and please see the [contributing guidelines](https://github.com/neetcode-gh/leetcode#contributing-guidelines) for details on what we're looking for in PRs!

PR #436, opened 2022-07-09
Update 242-Valid-Anagram.java

bshivaraj:


mitchellirvin (changes requested):
hi @bshivaraj ! thank you for contributing!

your change appears to be less efficient than the existing solution. the existing solution had a runtime of `O(s + 26)`, which reduces to `O(s)`, where `s` is the length of the input strings. your proposal is `O(s + s + (s * log(s)) + (s * log(s)) + s)`, which reduces to `O(s * log(s))`.

if you would like to add your solution as a separate approach, that would be okay! it's a very intuitive solution (sorting then comparing the arrays). see the [contributing guidelines](https://github.com/neetcode-gh/leetcode#contributing-guidelines) for details about best practices for getting your PRs reviewed and merged quickly

PR #444, opened 2022-07-10
Update 1-Two-Sum.java improved code readability

ColstonBod-oy:
Modified the way the variables are used and named as I find them a bit confusing at first glance.


java/1-Two-Sum.java
```diff
@@ -3,16 +3,16 @@ public int[] twoSum(int[] nums, int target) {
...
+            int diff = target - nums[i];
             
-            if (prevMap.containsKey(nums[i])) {
-                return new int[]{prevMap.get(num), i};
+            if (prevMap.containsKey(diffNum)) {
+                return new int[]{prevMap.get(diffNum), i};
             }
             
-            prevMap.put(target - num, i);
+            prevMap.put(diff, i);
```
mitchellirvin, line 13: this line change makes sense!

mitchellirvin (changes requested):
hi @ColstonBod-oy ! thank you for the contribution!

i'm not sure about the name `diffNum`. it implies "difference number", but that's not actually what it is (it's not the result of subtraction operation). it is the "current number in a loop over a collection of numbers". also, `diffNum` now clashes with `diff`, and it's not clear looking at them side by side what the difference is.

it's pretty standard to use the convention: `int type = types[i];` when capturing the current element in a loop, so i think the existing code of `int num = nums[i];` makes sense. you usually only see a more specific name, maybe `currentNum` or the like when there is a "next" or "previous" to track as well. 

i think the change on line 13 is a positive one, but in general I don't think the other changes improve the readability.

java/1-Two-Sum.java
ColstonBod-oy: @mitchellirvin I see, I was kinda hesitant of the new variable name as well, thanks for clarifying!

ColstonBod-oy:
Changed the code on line 13 to use the variable ```diff``` instead.

PR #447, opened 2022-07-10
refactor: js - trie

aakhtar3:


javascript/695-max-area-of-island.js
```diff
@@ -2,32 +2,35 @@
  * @param {number[][]} grid
  * @return {number}
  */
- var maxAreaOfIsland = function(grid) {
```
aakhtar3, line 5: This file has a conflicting issue due to main have 2 files with lowercase and uppercase

Ahmad-A0:
Thanks, I've removed the duplicate (`695-max-area-of-island.js`) from the main branch, that also means that one of the files you've committed is a duplicate so I'll remove that before merging.

Ahmad-A0:
Thanks, @aakhtar3!

PR #452, opened 2022-07-11
Adds another method to check for palindrome

andrewcar:


mitchellirvin (approved):
thanks for the contribution @andrewcar !

is it possible in swift to simply `return filteredString == String(filteredString.reversed())`?

i'm not familiar with swift, but it seems like it could be more readable this way? i'm down to merge either way just let me know what you prefer

PR #454, opened 2022-07-11
refactor: js - slidewindow

aakhtar3:


javascript/567-Permutation-In-String.js
```diff
@@ -1,158 +1,68 @@
...
-    const s2Chars = Object.create(null);
-    
-    for (const ch of s1) {
-        if (!(ch in s1Chars)) {
-            s1Chars[ch] = 0;
-            s2Chars[ch] = 0;
-        }
-        ++s1Chars[ch];
-    }
+var checkInclusion = (s1, s2) => {
```
mitchellirvin, line 7: i love the readability changes 💯 

i'd also love to keep both of the former solutions (because each is instructive in a helpful way) included. do you mind keeping both, and keeping the parts of the comments that explain how they're distinct?

javascript/76-Minimum-Window-Substring.js
```diff
@@ -1,35 +1,78 @@
...
+};
+
+const getWindowPointers = (s, t, frequencyMap) => {
+    let [ left, right, matched, start, end ] = [ 0, 0, 0, 0, (s.length + 1) ];
+    
+    while (right < s.length) {
+        matched = addRightFrequency(s, right, frequencyMap, matched);
+
+        const canSlide = () => matched === t.length;
+        while (canSlide()) {
```
mitchellirvin, line 32: nit: prefer boolean to lambda here

javascript/76-Minimum-Window-Substring.js
```diff
@@ -1,35 +1,78 @@
...
+    return matched;
+};
+
+const subtractLeftFrequency = (s, left, frequencyMap, matched) => {
+    const char = s[left];
+
+    if (frequencyMap.has(char)) {
+        const isOutOfWindow = frequencyMap.get(char) === 0;
+
+        matched -= Number(isOutOfWindow);
```
mitchellirvin, line 69: for readability's sake, converting a boolean to a number feels potentially confusing. maybe prefer:
```
if (isOutOfWindow) {
  matched--
}
```

mitchellirvin (changes requested):
@aakhtar3 thanks for contributing! this looks much better and really helps folks understand these solutions. 

there's one necessary change request on my part (to not remove the distinct solution in 567 and to keep the content of the comments), and a few optional nits that you can change if you'd like

javascript/76-Minimum-Window-Substring.js
aakhtar3: @mitchellirvin I mentioned the same issue in my other PR. Won't be able to a const in a while loop.

mitchellirvin:
thank you @aakhtar3 !

PR #459, opened 2022-07-11
Update JavaScript code for Balanced Binary Tree

YashSingh15:
The JavaScript solution for 110 - Balanced Binary Tree currently
available on the NeetCode.io website is inefficient, with a time
complexity of O(n^2). I have updated the solution to reflect the
original Python solution as presented in the video.


Ahmad-A0:
Thanks, @YashSingh15!

PR #465, opened 2022-07-12
Rust solutions

barlingo:
Moved rust solution to /rust folder.
Left previous solutions.


tomijaga:
Hey @barlingo,
I have reviewed this PR and have a few comments:
- **contains_duplicates**:   Your solutions optimises for space. While the neetcode video solutions optimise for speed by using a `hashmap` instead of a `vec`.

- **group_anagrams**: Good solution. One optimisation you could do is to store the key array directly in the hashmap instead of converting it to a `vec`.

- **top_k_frequent_elements**: Brilliant solution, sorting the elements into buckets, but this is different from the neetcode video solution, which uses quickselect.

- **two_sum**: Less optimal solution. Instead of a vec to store the results, you could use a HashSet. The reason for this is the`.contains()` method on a vec runs in `O(n)` time while in a HashSet, it runs at constant time ( `O(1)`). 

Thanks for your contributions. However, I will not be merging this PR because of these reasons.

PR #467, opened 2022-07-12
Update 242-Valid-Anagram.js

chaitanyax:
we should use hasOwnproperty because for value zero case if condition fails


Ahmad-A0:
Thanks, @chaitanyax!

PR #468, opened 2022-07-12
Swift solutions for 43, 50 and 121

seungjun-green:


Ahmad-A0:
Thanks, @seungjun-green! Sorry for the delay.

PR #481, opened 2022-07-13
refactored twoSum to make use of previously defined "diff" variable t…

Frogers98:
…hat was otherwise unused.


Frogers98:
already resolved by another pull request

PR #482, opened 2022-07-13
Update 543-Diameter-of-Binary-Tree.py

MeViMo:
The original `return -1` and `2+` actually cancel each other out and thus serve no purpose, only causing confusion.


AnandBaburajan:
@MeViMo nice! What do you think about changing the res variable from a `list` to an `int`, and making it `nonlocal` inside the `dfs` function? I found it confusing and was going to open a PR but decided to comment here instead since you already have a PR for the same problem.

```python
# Definition for a binary tree node.
# class TreeNode:
#     def __init__(self, val=0, left=None, right=None):
#         self.val = val
#         self.left = left
#         self.right = right
class Solution:
    def diameterOfBinaryTree(self, root: Optional[TreeNode]) -> int:
        res = 0
        
        def dfs(root: Optional[TreeNode]):
            nonlocal res

            if not root:
                return 0
            
            left_diameter = dfs(root.left)
            right_diameter = dfs(root.right)
            
            res = max(res, left_diameter + right_diameter)
            
            return 1 + max(left_diameter, right_diameter)
        
        dfs(root)
        
        return res
```

AnandBaburajan:
@MeViMo nice! What do you think about changing the res variable from a `list` to an `int`, and making it `nonlocal` inside the `dfs` function? I found it confusing and was going to open a PR but decided to comment here instead since you already have a PR for the same problem.

```python
# Definition for a binary tree node.
# class TreeNode:
#     def __init__(self, val=0, left=None, right=None):
#         self.val = val
#         self.left = left
#         self.right = right
class Solution:
    def diameterOfBinaryTree(self, root: Optional[TreeNode]) -> int:
        res = 0
        
        def dfs(root: Optional[TreeNode]):
            nonlocal res

            if not root:
                return 0
            
            left = dfs(root.left)
            right = dfs(root.right)
            
            res = max(res, left + right)
            
            return 1 + max(left, right)
        
        dfs(root)
        
        return res

```

MeViMo:
@AnandBaburajan That's probably what I would do to. You can push that change to this PR if you like. Up to the maintainers whether or not to include it.

AnandBaburajan:
I can push the changes but I think you need to resolve the conflicts first.

MeViMo:
I don't need to but I did for your convenience.

Ahmad-A0:
Thanks, @MeViMo and @AnandBaburajan!

PR #485, opened 2022-07-13
473-matchsticks to square

hnlCaseSensitive:
Sorry for uploading a duplicate last time! if there is a better implementation or one already existing, altho i did double check this time, feel free to delete.


Ahmad-A0:
No problem. Thanks, @hnlCaseSensitive!

PR #498, opened 2022-07-15
Added 2 sum solution using go

hnlCaseSensitive:


Ahmad-A0:
Sorry for the delay, unfortunately this solution is now a duplicate.

PR #505, opened 2022-07-16

chaitanyax:
Added more simple version of JavaScript solution


Ahmad-A0:
Sorry, since this solution uses sorting it doesn't fit the constraint that _"Your algorithm's time complexity must be better than **O(n log n),** where n is the array's size."_.

PR #507, opened 2022-07-16
994-Rotting-oranges JS solution

mariegobran:


Ahmad-A0:
Thanks, @mariegobran!

PR #513, opened 2022-07-17
over flow in one test-case

The-DefaultCube:
[805306368,805306368,805306368]
1000000000


Ahmad-A0:
Thanks, @The-DefaultCube!

PR #520, opened 2022-07-18
Update 746-Min-Cost-Climbing-Stairs.py

MeViMo:
`cost.append(0)` serves no purpose but to confuse.


Ahmad-A0:
Thanks, @MeViMo! Sorry for the delay!

PR #523, opened 2022-07-18
go: group anagrams

Aceix:
Add Go solution for [Group Anagrams](https://leetcode.com/problems/group-anagrams/)


Ahmad-A0:
Thanks, @Aceix!

PR #527, opened 2022-07-18
go solution for 217 contains duplicate

Damans227:
Adding go solution for problem #217: Contains Duplicate


Damans227:
Will create separate branches for solutions.

PR #551, opened 2022-07-20
Update min_stack.cpp

hanlinm2:
Made solution simpler and easier to understand by making minStk a stack of int instead of stack<pair<int,int>>. Passes all test cases on leetcode.


Ahmad-A0:
Sorry for the delay, @hanlinm2! Closing this as stale as the C++ files have moved.

PR #558, opened 2022-07-21
Update 853-Car-Fleet.py

UWDavidWu:
Reverse is a parameter in the sorted() function and setting it to be true is equivalent to sorted()[::-1] but more intuitive


Ahmad-A0:
Thanks, @UWDavidWu!

PR #561, opened 2022-07-21
Update 347-Top-k-frequent-elements.py

jayeshmutha221b:
This approach is much easy to understand


Ahmad-A0:
This approach is `O(nlogn)` as opposed to the previous `O(n)` solution, I think it might be better to add both rather than one or the other.

PR #563, opened 2022-07-21
Create 572-Subtree-of-Another-Tree.cs

notauserx:
add csharp solution for leetcode 572. Subtree of Another Tree


Ahmad-A0:
Thanks @notauserx!

notauserx:
My pleasure @Ahmad-A0!

PR #600, opened 2022-07-23
removing root level python sols in favor of the python/ dir

mitchellirvin:
as a final step in the effort to clean up the root directory, delete all the python solutions in root.

these all already exist in the `python/` dir. ideally this happens ASAP to help avoid divergence with open/new PRs. 

@neetcode-gh i'll let you handle merging this one because this will break links on the website, and you can make sure that things aren't broken for long!


neetcode-gh:
Thanks for doing this, I'll merge it tomorrow after I update the links!

DanMossa:
@neetcode-gh @mitchellirvin Love the new look and updates!

I am having an issue with this update though. I do like to look at the python and js code. Sometimes they're different and/or sometimes the javascript version does not exist. 

This change makes it so that I now have to go back to the top and constantly switch between the two languages. 

Any ideas on potentially adding an `all` option for the visible languages?

Once again, I'm super appreciate of all the hard work that's been put into this site!

mitchellirvin:
hi @DanMossa ! i think that type of problem is maybe best solved by browser features (e.g. having two separate browser windows side by side), as opposed to trying to maintain duplicate solutions in different directory structures? I may be misunderstanding the problem, though!

if you type `t` in github, you can do a fast file search for the file you're looking for (if that helps at all!)

PR #693, opened 2022-07-30
Java: 846. Hand of Straights

MaratKhakim:

- **File(s) Modified**: _846-Hand-of-Straights.java_
- **Language(s) Used**: _Java._
- **Submission URL**: _https://leetcode.com/submissions/detail/303536529/_

[//]: # "Finally copy the URL from the nav bar, it should look like https://leetcode.com/submissions/detail/xxxxxxxxx/"

### DIFF
--- a/java/846-Hand-of-Straights.java
+++ b/java/846-Hand-of-Straights.java
@@ -1,0 +1,31 @@
+class Solution {
+    public boolean isNStraightHand(int[] hand, int groupSize) {
+        if (hand.length % groupSize != 0)
+            return false;
+
+        TreeMap<Integer, Integer> map = new TreeMap<>();
+
+        for (int card : hand) {
+            map.put(card, map.getOrDefault(card, 0)+1);
+        }
+
+        while (!map.isEmpty()) {
+            int num = map.firstKey();
+            int k = groupSize;
+
+            while (map.containsKey(num) && k-- > 0) {
+                map.put(num, map.get(num)-1);
+
+                if (map.get(num) == 0)
+                    map.remove(num);
+
+                num++;
+            }
+
+            if (k > 0)
+                return false;
+        }
+
+        return true;
+    }
+}


mitchellirvin:
hi @MaratKhakim ! thanks so much for contributing. 

https://github.com/neetcode-gh/leetcode/pull/683 appears to solve the same problem in Java, and because we're treating PRs as first come first serve, we unfortunately won't be able to accept this solution because it'll be a duplicate after the other PR merges

an unrelated note: is this solution optimal in terms of runtime? every operation on the treemap (`get`, `put`, `containsKey`, `remove`) is `log (n)` runtime, where `n` is the number of keys in the map, and some of those operations are happening inside doubly nested loops in the code. I'd be curious to actually benchmark this approach against the priority queue approach that neet talks through in his video: https://www.youtube.com/watch?v=amnrMCVd2YI

MaratKhakim:
> hi @MaratKhakim ! thanks so much for contributing.
> 
> #683 appears to solve the same problem in Java, and because we're treating PRs as first come first serve, we unfortunately won't be able to accept this solution because it'll be a duplicate after the other PR merges
> 
> an unrelated note: is this solution optimal in terms of runtime? every operation on the treemap (`get`, `put`, `containsKey`, `remove`) is `log (n)` runtime, where `n` is the number of keys in the map, and some of those operations are happening inside doubly nested loops in the code. I'd be curious to actually benchmark this approach against the priority queue approach that neet talks through in his video: https://www.youtube.com/watch?v=amnrMCVd2YI

Thank you for the update and for your question.

Time complexity is O(M x min(logM, K) x logM), where M - number of unique elements, K - group size. The worst case scenario (when all elements are unique and not consecutive and K = N), then M = N and time complexity is O(Nx (logN)^2). But I think this solution would be faster in the average cases. It is interesting to see the benchmark of two solutions.

PR #854, opened 2022-08-19
416-Partition-Equal-Subset-Sum: remove unnecessary variable

Maximilian5189:

- **File(s) Modified**: 416-Partition-Equal-Subset-Sum.js
- **Language(s) Used**: javascript
- **Submission URL**: https://leetcode.com/submissions/detail/777616785/

Hey guys, the JS solution of partition equal subset sum has an variable which is not needed. Setting and checking the variable makes the code unnecessary complex to read. With removing the variable, the solution is closer to the original Python solution.


Ahmad-A0:
Thanks, @Maximilian5189!

PR #1172, opened 2022-09-23
Update 125-Valid-Palindrome.js

delsint123:
The first original solution seems to be O(n^2) since there is a nested loop in the isValid function. 

I have added a second O(n) solution in both time and space complexity.

- **File(s) Modified**: 125-Valid-Palindrome.js
- **Language(s) Used**: javascript
- **Submission URL**: https://leetcode.com/submissions/detail/806984435/

### DIFF
--- a/javascript/125-Valid-Palindrome.js
+++ b/javascript/125-Valid-Palindrome.js
@@ -1,6 +1,6 @@
 /**
  * https://leetcode.com/problems/valid-palindrome/
- * Time O(N) | Space O(1)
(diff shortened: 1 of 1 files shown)


aakhtar3:
@delsint123 Review my changes.
At first, a nested loop might look O(N^2), but for this problem that is not the case.
Since we are using 2 pointer algorithm, we are telling the second while loop to update the pointer.

aakhtar3 (approved):

delsint123:
Thanks for the clarification.

PR #1260, opened 2022-10-13
Modified code as in the video

elcabalero:
- **File(s) Modified**: cheapest_flights_within_k_stops.cpp
- **Language(s) Used**: C++
- **Submission URL**: https://leetcode.com/problems/cheapest-flights-within-k-stops/submissions/821560839/


Ahmad-A0 (dismissed):
Thanks for the optimization @elcabalero, if possible could you move the old solution below and renaming the `Solution` class, so that both approaches are visible.

Ahmad-A0:
Thanks for contributing!

PR #1265, opened 2022-10-15
Update search_a_2d_matrix.cpp

Abusagit:
Hello there, I\`d like to propose my own solution for the problem "74. Search a 2D Matrix" as it reduces the problem to "704. Binary Search" problem. Reduction is performed by considering matrix as whole 1D array of length `m * n`, elements in this "array" are stored in sorted order as well: access to the position inside a matrix can be done by 2 simple computations:
`Array[k] == Matrix[k / n][k % n]`
where n is the number of columns in a matrix. We simply linearize our matrix by computing middle elements of the array of length m*n and then bijectively access elements in a matrix by using formula above.

- **File(s) Modified**:  cpp/neetcode_150/05_binary_search/search_a_2d_matrix.cpp
- **Language(s) Used**: c++
- **Submission URL**: [https://leetcode.com/submissions/detail/822945700/](https://leetcode.com/submissions/detail/822945700/)

### DIFF
--- a/cpp/neetcode_150/05_binary_search/search_a_2d_matrix.cpp
+++ b/cpp/neetcode_150/05_binary_search/search_a_2d_matrix.cpp
@@ -47,3 +47,41 @@ public:
         return false;
     }
 };
+
+// Same approach with implicit array linearization
+// T(n) = O(log_2(mn)) = O(log_2(m) + log_2(n))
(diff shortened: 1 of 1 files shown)


Ahmad-A0:
Thanks for contributing!

PR #1286, opened 2022-10-19
Solution for 392-Is-Subsequence leetcode question in pyton

hemansah:
I have added a solution in python for Is-Subsequence leetcode question.


Ahmad-A0:
Sorry, this appears to be (a duplicate)[https://github.com/neetcode-gh/leetcode/blob/main/python/392-Is-Subsequence.py]

PR #1422, opened 2022-11-05
Move python/1641-Count-Sorted-Vowel-Strings.cpp to cpp/

i13e:
- **File(s) Modified**: python/1641-Count-Sorted-Vowel-Strings.cpp
- **Language(s) Used**: n/a
- **Submission URL**: n/a


Ahmad-A0:
Thanks, @i13e!

PR #1597, opened 2022-12-25
Create 26-Remove-Duplicates-from-Sorted-Array.js

aadil42:
Solved remove-duplicates-from-sorted-array with JS.

File(s) Added: 26-Remove-Duplicates-from-Sorted-Array.js
Language(s) Used: JavaScript
Submission URL: https://leetcode.com/problems/remove-duplicates-from-sorted-array/submissions/865213761/

### DIFF
--- a/javascript/26-Remove-Duplicates-from-Sorted-Array.js
+++ b/javascript/26-Remove-Duplicates-from-Sorted-Array.js
@@ -5,21 +5,20 @@
  * @param {number[]} nums
  * @return {number}
  */
-var removeDuplicates = function(nums) {
+var removeDuplicates = (nums) => {
+    let [left, right] = [0, 0];
 
-    let left = 0;
-    let right = 1;
-    let delCount = 0;
+    while (right < nums.length) {
+        const [leftVal, rightVal] = [nums[left], nums[right]];
 
-    while (right <= nums.length) {
-        if (nums[left] === nums[right]) {
-            right++;
-            delCount++;
-        } else {
-            nums.splice(left + 1, delCount);
-            delCount = 0;
+        const isEqual = (rightVal === leftVal);
+        if (!isEqual) {
             left++;
-            right = left + 1;
+            nums[left] = rightVal;
         }
+
+        right++;
     }
+
+    return (left + 1);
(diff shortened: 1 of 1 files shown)


javascript/26-Remove-Duplicates-from-Sorted-Array.js
```diff
@@ -0,0 +1,25 @@
...
+    let left = 0;
+    let right = 1;
+    let delCount = 0;
+
+    while (right <= nums.length) {
+        if (nums[left] === nums[right]) {
+            right++;
+            delCount++;
+        } else {
+            nums.splice(left + 1, delCount);
```
aakhtar3, line 19: This makes it N^2

aakhtar3 (changes requested):

javascript/26-Remove-Duplicates-from-Sorted-Array.js
aadil42: Because of recording the index of each element? for example, [1,2,3,3,4,5] if we remove 3 from the 3rd index then all the elements' index will be updated after index 3. Is this the reason? If so can you suggest a better approach for removing elements?

aakhtar3:
```js

/**
 * Two Pointer
 * Time O(N) | Space O(1)
 * https://leetcode.com/problems/remove-duplicates-from-sorted-array/
 * @param {number[]} nums
 * @return {number}
 */
var removeDuplicates = function(nums) {
    let [ left, right ] = [ 0, 0 ];
    
    while (right < nums.length) {/* Time O(N) */
        const [ leftVal, rightVal ] = [ nums[left], nums[right] ];
        const isEqual = (rightVal === leftVal);
        
        if (!isEqual) {
            left++;
            nums[left] = rightVal;
        }
        
        right++;
    }
    
    return (left + 1);
};

```

aadil42: That is really good. I updated the code please check.

aakhtar3 (approved):

PR #1731, opened 2022-12-29
Create 1905-count-sub-islands.java

AP-Repositories:
This java translation follows the python solution as closely as possible. There are 2 changes which have been made to make the python solution work in java. The first change is that the class RecursiveBiFunction is defined. This class exists so that the dfs lambda expression may refer to itself. The second change is that the visit set contains flattened integer coordinates instead of tuples. This change is made because tuples do not exist in java, and using an int array would not provide the correct behavior as int arrays compare themselves by not by value, but by reference.

- **File(s) Modified**: _1905-count-sub-islands.java_
- **Language(s) Used**: _java_
- **Submission URL**: _https://leetcode.com/submissions/detail/867724587/_

### DIFF
--- a/java/1905-count-sub-islands.java
+++ b/java/1905-count-sub-islands.java
@@ -1,0 +1,42 @@
+class Solution {
+    class RecursiveBiFunction<A, B, C> {
+        BiFunction<A, B, C> func;
+    }
+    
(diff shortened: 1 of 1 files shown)


Ahmad-A0:
Thanks, @AP-Repositories!

PR #2292, opened 2023-02-25
Delete newer duplicate of Problem 665 in C++ and rename older contribution

a93a:
There was a duplicate of 0665-non-decreasing-array.cpp. I deleted the older file, and rename the earlier file to follow the filename style. The older contribution was also in style with the video style, the newer wasn't.

- **File(s) Modified**: _0665-non-decreasing-array.cpp._
- **Language(s) Used**: _C++_
- **Submission URL**: _NO URL_


a93a:
Do NOT merge, let @neetcode-gh check to make sure that it's ok. Thanks.

PR #2296, opened 2023-02-26
Target sum (runtime and memory > 80%) #1

purnawirman:
File(s) Modified: 0494-target-sum.py
Language(s) Used: _python
Submission URL: https://leetcode.com/problems/target-sum/submissions/905306462/

optimize more by:
1. using only 2 rows of dp
2. iterate only on non zero previous row dp

### DIFF
--- a/python/0494-target-sum.py
+++ b/python/0494-target-sum.py
@@ -1,16 +1,26 @@
 class Solution:
+    """
+    Hello... first time contributor. Just used your resource, and found it super helpful. 
+    I noticed you have habit of running DP problem from backward (e.g. for i in range(n-1, -1, -1, instead of for in in range(n), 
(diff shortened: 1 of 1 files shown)


raul-sauco:
Same idea but using a Counter makes it a bit more readable, in my opinion, and the performance is similar. 

Runtime 304 ms Beats 87.41%
Memory 14.1 MB Beats 88.17%

https://leetcode.com/problems/target-sum/submissions/842843238/

```
class Solution:
    def findTargetSumWays(self, nums: List[int], target: int) -> int:
        dp = Counter([0])
        for num in nums:
            next = Counter()
            for key in dp.keys():
                next[key - num] += dp[key]
                next[key + num] += dp[key]
            dp = next
        return dp[target]
```

PR #2345, opened 2023-03-11
Update 0200-number-of-islands.ts

M0Leo:
- **File(s) Modified**: 0200-number-of-islands.ts
- **Language(s) Used**: TypeScript
- **Submission URL**: https://leetcode.com/problems/number-of-islands/


Ykhan799:
@M0Leo Instead of deleting the BFS solution, you should add your DFS solution below.

PR #2445, opened 2023-05-01
Update 0046-permutations.kt

FilipeLipan:
Made a solution closer to the python solution, the previous kotlin solution it's faster, but very different from the video explanation https://www.youtube.com/watch?v=s7AvT7cGdSo
 
 * **File(s) Modified**: 0046-permutations.kt
 * **Language(s) Used**: kotlin
 * **Submission URL**: https://leetcode.com/problems/permutations/submissions/942342203/

### DIFF
--- a/kotlin/0046-permutations.kt
+++ b/kotlin/0046-permutations.kt
@@ -1,3 +1,30 @@
+// original solution
+class Solution {
+    fun permute(nums: IntArray): List<List<Int>> {
+        val res = mutableListOf<List<Int>>()
+        permute(nums, mutableSetOf<Int>(), mutableListOf<Int>(), res)
+        return res
+    }
+
+    fun permute(nums: IntArray, set: MutableSet<Int>, list: MutableList<Int>, res: MutableList<List<Int>>) {
+        if (list.size == nums.size) {
+            res.add(ArrayList(list))
+            return
+        }
+
+        for (i in 0..nums.size-1) {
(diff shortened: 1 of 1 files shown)


a93a:
Hi! Your contribution link seems to be incorrect. Could you provide the correct one, please ?

a93a:
@FilipeLipan Thanks! I don't see a reason to completly overrid the old solution, but we can add yours too (several different solutions in the same file are OK, as long as they differ but still manage to solve the problem).

Could you do some changes to your file and add some separation between the solutions, and I could merge.

for example:

```
// solution one
class Solution {
    fun permute(nums: IntArray): List<List<Int>> {
        //....
    }
}

// solution two 
class Solution {
    fun permute(nums: IntArray): List<List<Int>> {
        //....
    }
}
```

FilipeLipan:
@a93a, I've done it! I'm submitting this solution because I found it annoying to study with the other solution that had a different strategy from the one in the video (https://www.youtube.com/watch?v=s7AvT7cGdSo). Please let me know if you have any questions or change requests.

a93a:
Yes I understand, thanks for the help. You can put the video solution at top instead if you would like, I think it's better? So it's the first you see when you check the file.

FilipeLipan:
@a93a done!!

a93a:
Thanks a lot @FilipeLipan and thanks for your contribution!

PR #2451, opened 2023-05-04
Fix: Big O in 0235-lowest-common-ancestor-of-a-binary-search-tree.js

pobch:
- **File(s) Modified**: _0235-lowest-common-ancestor-of-a-binary-search-tree.js_
- **Language(s) Used**: _javascript_

### Description
The time complexity of the current solution should be O(H), **not** O(N) when N is the total number of node and H is the height of the tree (in a specific case H will equal to N).


Ykhan799:
@pobch Thanks for the correction, will merge this change in

PR #2455, opened 2023-05-05
Skip usage of the 'visit' set. All you need is just add an extra condition

RomanGirin:
You can skip usage of the 'visit' set. All you need is just add a condition 'if rooms[r][c] >= dist:' in line 34. It reduces memory usage as well as makes the code shorter.


a93a:
Can you provide a submission URL with this change, please

RomanGirin:
> 

It's locked on leetcode problem, so I can't, sorry

PR #2473, opened 2023-05-13
Created: 0278-first-bad-version.py

noamtamir:
- **File(s) Modified**: _0278-first-bad-version.py_
- **Submission URL**: _https://leetcode.com/problems/first-bad-version/submissions/949592817/_


Ykhan799:
This solution is similar to the existing solution.

PR #2474, opened 2023-05-13
fix and simplify rust solution for `0036-valid-sudoku`

Adimac93:
- **File(s) Modified**: _0036-valid-sudoku.rs_
- **Language(s) Used**: _rust_
- **Submission URL**: _https://leetcode.com/problems/valid-sudoku/submissions/949636309/_

Previous incorrect submission: _https://leetcode.com/problems/valid-sudoku/submissions/949595521/_

1. fix for `bx` indices 
2. use of returned insertion status 
 
> **Note**
> _https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert_
> If the set did not previously contain this value, `true` is returned.
> If the set already contained this value, `false` is returned.

### DIFF
--- a/rust/0036-valid-sudoku.rs
+++ b/rust/0036-valid-sudoku.rs
@@ -12,28 +12,22 @@ impl Solution {
             for j in 0..9{
                 let r = board[i][j];
                 let c = board[j][i];
-                let b = board[i / 3 * 3 + j / 3][i/3 * 3 + j%3];
+                let b = board[i / 3 * 3 + j / 3][i % 3 * 3 + j % 3];
                 
                 if r != '.'{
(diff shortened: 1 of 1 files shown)


oliverocean:
@Adimac93 thanks so much for posting this... I was banging on the current "solution" for several days, trying to figure out what was wrong... was about to give up and then found your PR that makes it work correctly! :)

oliverocean:
@neetcode-gh please merge this in if possible (and put on neetcode.io)! The current Rust solution has a small but critical error, the PR here by @Adimac93 fixes the solution and also makes it more idiomatic. Thanks!

tahsintunan (approved):

PR #2503, opened 2023-05-22
Create 0343-integer-break.py

pdequina:
File(s) Modified: 0343-integer-break.py
Language(s) Used: python
Submission URL: https://leetcode.com/problems/integer-break/submissions/955194149/

### DIFF
--- a/python/0343-integer-break.py
+++ b/python/0343-integer-break.py
@@ -1,0 +1,15 @@
+class Solution(object):
+    def integerBreak(self, n):
+        """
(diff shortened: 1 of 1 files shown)


neetcode-gh:
Is there a reason you're using python2 instead of python3? 

Also, while your solution is correct, I'd like to keep it similar to the code presented in the videos, since sometimes users complain that the code is different from the video.

After a few more PRs feel free to tag me and i'll grant you collab permissions.

PR #2522, opened 2023-05-28
Cleaner code for 0225-implement-stack-using-queues.py

aroch17:
- **File Modified**:  0225-implement-stack-using-queues.py
- **Language Used**: python
- **Submission URL**: https://leetcode.com/problems/implement-stack-using-queues/submissions/959236812/

I have modified the python solution for this problem as in the original solution the code
```  
for _ in range(len(self.q) - 1):
    self.q.push(self.q.popleft())
``` 
was reused in two different functions - pop() and top() to be specific. 

We can eliminate this repetition by reversing the order of the queue in the push() method itself. This makes the code more readable and easier to understand. The time and space complexities remain the same i.e O(n).

Hope this helps!

### DIFF
--- a/python/0225-implement-stack-using-queues.py
+++ b/python/0225-implement-stack-using-queues.py
@@ -4,18 +4,15 @@ class MyStack:
 
     def push(self, x: int) -> None:
         self.q.append(x)
+        
+        for _ in range(len(self.q) - 1):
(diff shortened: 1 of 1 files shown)


Ykhan799:
@aroch17 Good refactoring. Thanks for your contribution 👍

PR #2544, opened 2023-06-07
Create 0912-sort-an-array.js

aadil42:
Added solution for Sorting an array in JS.

File(s) Added: 0912-sort-an-array.js
Language(s) Used: JavaScript
Submission URL: https://leetcode.com/problems/sort-an-array/submissions/984623618/

### DIFF
--- a/javascript/0912-sort-an-array.js
+++ b/javascript/0912-sort-an-array.js
@@ -1,0 +1,54 @@
+/**
+ * https://leetcode.com/problems/sort-an-array/
+ * Merge Sort
+ * Time O(n*log(n)) | Space O(n)
+ * @param {number[]} nums
+ * @return {number[]}
+ */
+var sortArray = function(nums) {
+    
+    const merge = (left, mid, right) => {
+        const arr1 = nums.slice(left, mid+1);
+        const arr2 = nums.slice(mid+1, right + 1);
+        
+        let p1 = 0; 
+        let p2 = 0;
+        let gp = left;
+
+        while(p1 < arr1.length && p2 < arr2.length) {
+            if(arr1[p1] < arr2[p2]) {
+                nums[gp] = arr1[p1];
+                p1++;
+            } else {
+                nums[gp] = arr2[p2];
+                p2++;
+            }
+            gp++;
+        }
+
+        while(p1 < arr1.length) {
+            nums[gp] = arr1[p1];
+            p1++;
+            gp++;
+        }
+
+        while(p2 < arr2.length) {
+            nums[gp] = arr2[p2];
+            p2++;
+            gp++;
+        }
+        return nums;
+    }  
+
+    const mergeSort = (left, right) => {
+
+        if(left === right) return nums;
+        
+        const mid = Math.floor((left+right)/2);
+        mergeSort(left, mid);
+        mergeSort(mid+1, right);
+        return merge(left, mid, right);
+    }
+
+    return mergeSort(0, nums.length  - 1);
+};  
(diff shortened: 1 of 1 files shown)


oldSheep017:
I got a easy one.

oldSheep017:
I got easier one, take a look
```js
return nums.sort((a, b) => a - b)
```

aadil42:
That is correct. Normally we do use the inbuilt sorting method. This was just for learning purposes.

javascript/0912-sort-an-array.js
```diff
@@ -0,0 +1,54 @@
+/**
+ * https://leetcode.com/problems/sort-an-array/
+ * Merge Sort
+ * Time O(n*log(n)) | Space O(n)
+ * @param {number[]} nums
+ * @return {number[]}
+ */
+var sortArray = function(nums) {
+    
+    const merge = (left, mid, right) => {
```
aakhtar3, line 10: Suggestion: Avoid using global scope with nested functions

aakhtar3 (changes requested):

javascript/0912-sort-an-array.js
aadil42: I don't understand. What do you mean. As far as I can understand You don't want me to use nums variable  inside the merge (nested function). But still I need to confirm this. The reason I have nested the merge function inside sortArray is because I don't want to pass the  nums veriable again and again.

aakhtar3: It’s ok to add as a new parameter to avoid using global scope.

```js
const sortArray = (nums, left = 0, right = (nums.length - 1)) =>
    mergeSort(nums, left, right);

const merge = (nums, left, right) => { … }

const sort = (nums, left, right, mid) => { … }

```

javascript/0912-sort-an-array.js
```diff
@@ -0,0 +1,54 @@
...
+            gp++;
+        }
+        return nums;
+    }  
+
+    const mergeSort = (left, right) => {
+
+        if(left === right) return nums;
+        
+        const mid = Math.floor((left+right)/2);
```
aakhtar3, line 47: Suggestion: Use Bitwise
```js
const mid = ((left + right) >> 1 );
```

javascript/0912-sort-an-array.js
aadil42: Hmm, I have created each function outside of the main function. Let me know if You would like to make further changes.

javascript/0912-sort-an-array.js
aadil42: Oh boy. I think it would be too arcane for people to understand. I usually go for this. mid = left + Math.floor((right  - left)/2); But for some reason, I'm not doing that here. Are you sure we should use bit manipulation here? It would be too difficult for people to grasp what we are doing here. That's just a personal opinion. Let me know if you still want me to proceed with the bit manipulation. :)

aakhtar3 (approved):

javascript/0912-sort-an-array.js
aakhtar3: @aadil42  I merged this pr before I got a chance to address this feedback.

The benefit of bitwise when doing binary search is to double dip and allow yourself to build connections of both patterns.

Bitwise is not a common interview problem, but when it is a problem you might forget how it works.
Using it in binary search will reinforce the programming mechanic. Binary search is meant to cut in half, and bitwise can do that.

aadil42: I think I need to think about it clearly. I never halved a number by using the bitwise operator. I will have to look into it. But thanks for the suggestion.

PR #2604, opened 2023-06-20
Create 1498-number-of-subsequences-that-satisfy-the-given-sum-conditi…

aadil42:
Solved number-of-subsequences-that-satisfy-the-given-sum-condition in JS.

File(s) Added: 1498-number-of-subsequences-that-satisfy-the-given-sum-condition.js
Language(s) Used: JavaScript
Submission URL: https://leetcode.com/problems/number-of-subsequences-that-satisfy-the-given-sum-condition/submissions/947074363/

### DIFF
--- a/javascript/1498-number-of-subsequences-that-satisfy-the-given-sum-condition.js
+++ b/javascript/1498-number-of-subsequences-that-satisfy-the-given-sum-condition.js
@@ -1,0 +1,30 @@
+/**
+ * Two Pointers
+ * https://leetcode.com/problems/number-of-subsequences-that-satisfy-the-given-sum-condition/
+ * Time O(n) | Space O(1)
+ * @param {number[]} nums
+ * @param {number} target
+ * @return {number}
+ */
+
+var numSubseq = function(nums, target) {
+    let total = 0n;
+    nums = nums.sort((a,b) => a-b);
+    const mod = BigInt(10**9 + 7);
+    
+    for(let i = 0; i < nums.length; i++) {
+        let left = i;
+        let right = i;
+        // if there's even a one element that exceeds target when adding it twice (min+max)  [5] 5+5 > 9 so we shouln't consider anything and just break because we won't find anything now.
+        if(nums[left]*2 > target) break;
+        while(right < nums.length && nums[left] + nums[right] <= target) {
+            right++;
+        }
+
+        // we are subtracting 1 from right because we would have gone over the subsequance by 1 // think about it.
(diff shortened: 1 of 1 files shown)


aakhtar3 (approved):

javascript/1498-number-of-subsequences-that-satisfy-the-given-sum-condition.js
```diff
@@ -0,0 +1,30 @@
...
+        let left = i;
+        let right = i;
+        // if there's even a one element that exceeds target when adding it twice (min+max)  [5] 5+5 > 9 so we shouln't consider anything and just break because we won't find anything now.
+        if(nums[left]*2 > target) break;
+        while(right < nums.length && nums[left] + nums[right] <= target) {
+            right++;
+        }
+
+        // we are subtracting 1 from right because we would have gone over the subsequance by 1 // think about it.
+        total += (BigInt(2)**BigInt(right - 1 - left));
```
aakhtar3, line 25: BigInt will increase Time complexity

aakhtar3 (changes requested):

javascript/1498-number-of-subsequences-that-satisfy-the-given-sum-condition.js
aadil42: I don't get how BigInt will increase the time complexity. I'm using BigInt because the value was overflowing. However, I realized that the time complexity should be n*log(n) because of the sorting. I updated that time complexity.

aadil42: Ok, so the Time complexity should be n*log(n) * (the time it takes to convert BigInt to a number which is O(n) if it's converted from the string). so it should be n*log(n)*k where k is equal to the time it takes to convert a BitInt to a Number right?

aadil42: @aakhtar3, any update on this?

aadil42: Waiting for an update...

aadil42: @aakhtar3, Is the reasoning correct for this one? Waiting for an update.

> Ok, so the Time complexity should be n_log(n) * (the time it takes to convert BigInt to a number which is O(n) if it's converted from the string). so it should be n_log(n)*k where k is equal to the time it takes to convert a BitInt to a Number right?

PR #2614, opened 2023-06-21
Create 0071-simplify-path.js

aadil42:
Solved simplify-path in JS.

File(s) Added: 0071-simplify-path.js
Language(s) Used: JavaScript
Submission URL: https://leetcode.com/problems/simplify-path/submissions/986179327/

### DIFF
--- a/javascript/0071-simplify-path.js
+++ b/javascript/0071-simplify-path.js
@@ -1,31 +1,36 @@
+
 /**
  * Stack
- * https://leetcode.com/problems/simplify-path/
- * 
- * Time O(n) | Space O(n)
+ * Time O(N) | Space O(N)
+ * https://leetcode.com/problems/simplify-path
  * @param {string} path
  * @return {string}
  */
-var simplifyPath = function(path) {
-    let currunt = '';
-    let myStack = [];
-    path = `/${path}/`;
-    for(let i = 0; i < path.length; i++) {
-        
-        if(path[i] === '/') {
-            if(currunt == '..') {
-                myStack.pop();
-            } else if(currunt !== '' && currunt !== '.') {
-                myStack.push(currunt);
-            }
-            currunt = '';
-        } else {
-            currunt += path[i];
-        }
-        
-    }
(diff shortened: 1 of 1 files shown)


javascript/0071-simplify-path.js
```diff
@@ -0,0 +1,33 @@
...
+ * https://leetcode.com/problems/simplify-path/
+ * 
+ * Time O(n) | Space O(n)
+ * @param {string} path
+ * @return {string}
+ */
+var simplifyPath = function(path) {
+    let currunt = '';
+    let myStack = [];
+    path = '/' + path + '/';
```
aakhtar3, line 12: Suggestion: Use interpolation

```js
path = `/${path}/`
```

javascript/0071-simplify-path.js
```diff
@@ -0,0 +1,33 @@
...
+    for(let i = 0; i < path.length; i++) {
+        
+        console.log(myStack);
+        if(path[i] === '/') {
```
aakhtar3, line 16: Suggestion: Avoid nesting more than 3 times

aakhtar3 (changes requested):

javascript/0071-simplify-path.js
aadil42: I don't see any other way. Can you kindly show me How can I avoid nested if statements and still get the logic right? Thank you :)

aakhtar3: use `continue;` when you need to exit early to avoid else

javascript/0071-simplify-path.js
aadil42: Actually. We can just remove that third if statement. if the stack is empty then nothing would happen. We don't even need to check it. I updated the code please have a look.

aadil42: yep, it's more readable. It's intimidating at first but if you try to read the code it's quite better than mine. Nicely Modularized.  I have updated the file. Please have a look.

aakhtar3 (approved):

PR #2621, opened 2023-06-22
Create 1898-maximum-number-of-removable-characters.js

aadil42:
Solved maximum-number-of-removable-characters in JS.

File(s) Added: 1898-maximum-number-of-removable-characters.js
Language(s) Used: JavaScript
Submission URL: https://leetcode.com/problems/maximum-number-of-removable-characters/submissions/977122834/

### DIFF
--- a/javascript/1898-maximum-number-of-removable-characters.js
+++ b/javascript/1898-maximum-number-of-removable-characters.js
@@ -1,0 +1,89 @@
+/**
+ * https://leetcode.com/problems/maximum-number-of-removable-characters/
+ * 
+ * Brute force
+ * Time O(removable.length * s.length) | Space O(1)
+ * @param {string} s
+ * @param {string} p
+ * @param {number[]} removable
+ * @return {number}
+ */
+var maximumRemovals1 = function(s, p, removable) {
+
+    let k = 0;
+    // removable.reverse();
+    s = s.split('');
+    p = p.split('');
+    for(let i = 0; i < removable.length; i++) {
+        s[removable[i]] = -1;
+        if(isSubSet()) {
+            k++;
+        } else {
+            return k;
+        }
+    }
+
+  function isSubSet() {
+      let i = 0;
+      let j = 0;
+
+      while(i < s.length && j < p.length) {
+          if(s[i] === p[j]) {
+              i++;
+              j++;
+          } else {
+              i++;
+          }
+      }
+      return j === p.length
+  }
+
+  return k;  
+};
+
+
+/**
+ * 
+ * Binary Search 
+ * n = length of string, k = length of removable
+ * Time O(log(k)*n) | Space O(k)
+ * @param {string} s
+ * @param {string} p
+ * @param {number[]} removable
+ * @return {number}
+ */
+var maximumRemovals = function(s, p, removable) {
+
+    s = s.split('');
+    p = p.split('');
+    let left = 0;
+    let right = removable.length - 1;
+    let k = 0;
+    while(left <= right) {
+        const mid = Math.floor((left+right)/2);
+        const hash = new Set(removable.slice(0, mid + 1));
+        if(isSubSet(hash))  {
+            k = Math.max(k, mid + 1);
+            left = mid + 1;
+        } else {
+            right = mid - 1;
+        }
+    }
+
+  function isSubSet(hash) {
+      let i = 0;
+      let j = 0;
+
+      while(i < s.length && j < p.length) {
+          if(s[i] === p[j] && !hash.has(i)) {
+              i++;
+              j++;
+          } else {
+              i++;
+          }
+      }
+      return j === p.length
+  }
+
+  return k;  
+};
--- a/javascript/1898-maximum-number-of-removable-characters.js
+++ b/javascript/1898-maximum-number-of-removable-characters.js
@@ -18,9 +18,9 @@ var maximumRemovals1 = function(s, p, removable) {
     s[removable[i]] = -1;
     if (isSubSet1(s, p)) {
       k++;
-    } else {
-      return k;
-    }
+      continue;
+    } 
+    return k;
   }
 
   return k;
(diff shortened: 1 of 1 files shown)


javascript/1898-maximum-number-of-removable-characters.js
```diff
@@ -0,0 +1,89 @@
...
+    for(let i = 0; i < removable.length; i++) {
+        s[removable[i]] = -1;
+        if(isSubSet()) {
+            k++;
+        } else {
+            return k;
+        }
+    }
+
+  function isSubSet() {
```
aakhtar3, line 26: Suggestion: Avoid internal functions that relay on global scope

javascript/1898-maximum-number-of-removable-characters.js
```diff
@@ -0,0 +1,89 @@
...
+
+    let k = 0;
+    // removable.reverse();
+    s = s.split('');
+    p = p.split('');
+    for(let i = 0; i < removable.length; i++) {
+        s[removable[i]] = -1;
+        if(isSubSet()) {
+            k++;
+        } else {
```
aakhtar3, line 21: Suggestion: Avoid else

```js
if (true) continue;

return ...;
```

aakhtar3 (changes requested):

javascript/1898-maximum-number-of-removable-characters.js
aadil42: done

javascript/1898-maximum-number-of-removable-characters.js
aadil42: done

javascript/1898-maximum-number-of-removable-characters.js
```diff
@@ -0,0 +1,96 @@
...
+ * Time O(removable.length * s.length) | Space O(1)
+ * @param {string} s
+ * @param {string} p
+ * @param {number[]} removable
+ * @return {number}
+ */
+var maximumRemovals1 = function(s, p, removable) {
+  let k = 0;
+  // removable.reverse();
+  s = s.split('');
```
aakhtar3, line 14: This will increase space complexity

javascript/1898-maximum-number-of-removable-characters.js
```diff
@@ -0,0 +1,96 @@
...
+  p = p.split('');
+
+  for (let i = 0; i < removable.length; i++) {
+    s[removable[i]] = -1;
+    if (isSubSet1(s, p)) {
+      k++;
+    } else {
```
aakhtar3, line 21: Avoid else

aakhtar3 (changes requested):

javascript/1898-maximum-number-of-removable-characters.js
aadil42: I updated the space complexity to n. But we already had s and p as strings we just used them in an array instead of a string. wouldn't time complexity be the same for that? Before we were using n space for string which we were getting as an argument. Now we use the same number of array elements.

In other words, since we're just converting some string of n char to be an array of n elements. Then, the space complexity is the same as before right? because we were already using space(n) for the strings. Let me know if I'm making a mistake. Thank you.

javascript/1898-maximum-number-of-removable-characters.js
aadil42: Done.

javascript/1898-maximum-number-of-removable-characters.js
aakhtar3: Sting is immutable primitive data type. You can not modify its state or space in memory.

If you split and convert to char array, it will need new space to allow the list to exist.

If you need to modify the string you can split, but if you need to read by index, you do not need to split and it will lower space complexity.

```js
/* Time O(N + N) | Space O(N)
for (const char of str.split('')) { ... }

/* Time O(N) | Space O(1)
for (const char of str) { ... }

/* Time O(N) | Space O(1)
for (const index char in str) {
	const char = str.charAt(index);
}
```

javascript/1898-maximum-number-of-removable-characters.js
aadil42: Sorry for the delay. I got busy. I have updated the space complexity as suggested. I have removed the _else_ case as well. Please check.

javascript/1898-maximum-number-of-removable-characters.js
aadil42: @aakhtar3, I updated the space complexity.

javascript/1898-maximum-number-of-removable-characters.js
aadil42: @aakhtar3, sorry for the delay. I got busy. I have removed the _else_ case as suggested. Please check.

javascript/1898-maximum-number-of-removable-characters.js
aadil42: @aakhtar3, It's done. I  removed the else case.

aakhtar3 (changes requested):

javascript/1898-maximum-number-of-removable-characters.js
aadil42: @aakhtar3, but if we don't convert it to the array then how are we going to update a specific char of the string? See line 18.
I did remove the split method on the binary-search solution and used the bitwise operation to find the mid now. Please check.

javascript/1898-maximum-number-of-removable-characters.js
```diff
@@ -0,0 +1,96 @@
...
+ */
+var maximumRemovals = function(s, p, removable) {
+  s = s.split('');
+  p = p.split('');
+  let left = 0;
+  let right = removable.length - 1;
+  let k = 0;
+
+  while (left <= right) {
+    const mid = Math.floor((left + right) / 2);
```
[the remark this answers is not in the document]
aadil42: It's done, using bitwise operation.

PR #2629, opened 2023-06-24
Create 2390-removing-stars-from-a-string.py

TheHong:
- **File(s) Modified**: 2390-removing-stars-from-a-string.py
- **Language(s) Used**: Python
- **Submission URL**: [https://leetcode.com/problems/removing-stars-from-a-string/submissions/978233431/](https://leetcode.com/problems/removing-stars-from-a-string/submissions/978233431/)


Ykhan799:
This solution is the same as the existing solution

PR #2737, opened 2023-07-24
Update: 0105-Construct-Binary-Tree-From-Preorder-And-Inorder-Traversal

dichlorodiphen:

- **File(s) Modified**: _0105-construct-binary-tree-from-preorder-and-inorder-traversal.java_
- **Language(s) Used**: _java_
- **Submission URL**: _https://leetcode.com/problems/construct-binary-tree-from-preorder-and-inorder-traversal/submissions/1003061292/_

### Important

### Changes

Modified solution to use sublists, which are backed by the original list, avoiding the linear-time copy from `copyOfRange`. New solution reads closer to the Python solution using slicing.

### DIFF
--- a/java/0105-construct-binary-tree-from-preorder-and-inorder-traversal.java
+++ b/java/0105-construct-binary-tree-from-preorder-and-inorder-traversal.java
@@ -14,32 +14,25 @@
  * }
  */
 class Solution {
-
     public TreeNode buildTree(int[] preorder, int[] inorder) {
-        if (preorder.length == 0 || inorder.length == 0) return null;
+        return build(Arrays.stream(preorder).boxed().toList(), Arrays.stream(inorder).boxed().toList());
+    }
 
-        TreeNode root = new TreeNode(preorder[0]);
-        int mid = 0;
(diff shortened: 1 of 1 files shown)


felivalencia3 (changes requested):
Your solution runs correctly, but it's much slower than the existing code in the repository. There is another way to solve this problem that is closer to Neetcode's original python solution but without the overhead that comes from using the streams and O(n) indexOf lookUps in your code.

felivalencia3:
Consider something like this solution.

```
class Solution {
    private int i = 0;
    private int p = 0;

    public TreeNode buildTree(int[] preorder, int[] inorder) {
        return build(preorder, inorder, Integer.MIN_VALUE);
    }

    private TreeNode build(int[] preorder, int[] inorder, int stop) {
        if (p >= preorder.length) {
            return null;
        }
        if (inorder[i] == stop) {
            ++i;
            return null;
        }

        TreeNode node = new TreeNode(preorder[p++]);
        node.left = build(preorder, inorder, node.val);
        node.right = build(preorder, inorder, stop);
        return node;
    }
}
```

PR #2743, opened 2023-07-28
Update 0049-group-anagrams.go

ctjnkns:
Uses the more efficient solution from the neetcode video (where you create an array of ints representing how many times each letter occurs in a string and use that as a key).


felivalencia3 (approved):
Great! Thank you for making the solution follow Neetcode's official solution.

PR #2827, opened 2023-08-11
Create 2485-find-the-pivot-integer.js

chetannada:

- **File(s) Modified**:  2485-find-the-pivot-integer.js
- **Language(s) Used**:  javascript
- **Submission URL**: https://leetcode.com/problems/find-the-pivot-integer/submissions/1015789860/

### Important

### DIFF
--- a/javascript/2485-find-the-pivot-integer.js
+++ b/javascript/2485-find-the-pivot-integer.js
@@ -1,0 +1,26 @@
+/**
+ * @param {number} n
+ * @return {number}
+ */
+var pivotInteger = function (n) {
+
+    // initialize function sum with parameter start and end to find sum between start and end and return num
+    function sum(start, end) {
+        let num = 0;
+        for (let i = start; i <= end; i++) {
+            num += i;
+        }
+        return num;
+    }
+
+    // loop through the 1 to n
+    for (let i = 1; i <= n; i++) {
+
+        // if sum of 1 to i and sum of i to n is equal then return i
+        if (sum(1, i) == sum(i, n)) {
+            return i;
+        }
+    }
+
+    return -1; // otherwise return -1
+};


ethancdaniel (changes requested):
There's a lot of repeated calculations in this implementation, I think modifying a few things in your approach (O(n) time) would be better. Instead of recalculating the prefix sum, we can progressively add it, and we don't have to do a special calculation for the right side, because we can get the right side's sum from `total - left side + x`.

```javascript
var pivotInteger = function(n) {
    let totalSum = (n * (n + 1)) / 2;
    let prefixSum = 0;

    for (let x = 1; x <= n; x++) {
        prefixSum += x;

        if (prefixSum === totalSum - prefixSum + x) {
            return x;
        }
    }

    return -1;
};
```
You could include this math-y implementation as well, but since it's harder to arrive to the answer, it probably shouldn't be listed as the main solution. O(1) runtime.

```javascript
		const sumToN= n * (n + 1) / 2;
        const sqrtVal = Math.sqrt(sumToN);
        return Math.floor(sqrtVal) === sqrtVal ? Math.floor(sqrtVal) : -1;
```

chetannada:
> There's a lot of repeated calculations in this implementation, I think modifying a few things in your approach (O(n) time) would be better. Instead of recalculating the prefix sum, we can progressively add it, and we don't have to do a special calculation for the right side, because we can get the right side's sum from `total - left side + x`.
> 
> ```js
> var pivotInteger = function(n) {
>     let totalSum = (n * (n + 1)) / 2;
>     let prefixSum = 0;
> 
>     for (let x = 1; x <= n; x++) {
>         prefixSum += x;
> 
>         if (prefixSum === totalSum - prefixSum + x) {
>             return x;
>         }
>     }
> 
>     return -1;
> };
> ```
> 
> You could include this math-y implementation as well, but since it's harder to arrive to the answer, it probably shouldn't be listed as the main solution. O(1) runtime.
> 
> ```js
> 		const sumToN= n * (n + 1) / 2;
>         const sqrtVal = Math.sqrt(sumToN);
>         return Math.floor(sqrtVal) === sqrtVal ? Math.floor(sqrtVal) : -1;
> ```

Hi @ethancdaniel 
Thanks for going deeply. I have added Optimize Solution as per your suggestion: Check it out - https://leetcode.com/problems/find-the-pivot-integer/submissions/1019969066/

felivalencia3 (approved):

PR #2960, opened 2023-09-01
Update 0567-permutation-in-string.py

coopers:
- **File(s) Modified**: _0567-permutation-in-string.py_
- **Submission URL**: _https://leetcode.com/problems/permutation-in-string/submissions/1037991382/_

Calculate matches in one-liner.
Only use a left pointer as the right point is always left + len(s1).


coopers:
Closing since it has previous 2 commits.

PR #2968, opened 2023-09-03
Update 0271-encode-and-decode-strings.py

coopers:
- **File(s) Modified**: _0271-encode-and-decode-strings.py_
- **Submission URL**: _https://leetcode.com/problems/encode-and-decode-strings/submissions/1039753932/_

Refactor to use f-string in encode function.


MHamiid (approved):
Thanks.

PR #3027, opened 2023-09-28
Update 0703-kth-largest-element-in-a-stream.py

coopers:
- **File(s) Modified**: _0703-kth-largest-element-in-a-stream.py_
- **Submission URL**: _https://leetcode.com/problems/kth-largest-element-in-a-stream/submissions/1061780994/_


neetcode-gh:
I think we should leave the code as-is to keep it consistent with the video

PR #3093, opened 2023-10-22
Update 0004-median-of-two-sorted-arrays.rs

emanuele-em:

- **File(s) Modified**: 0004-median-of-two-sorted-arrays.rs
- **Language(s) Used**: rust
- **Submission URL**: [_https://leetcode.com/problems/[problem-name]/submissions/1081164139/_](https://leetcode.com/problems/median-of-two-sorted-arrays/submissions/1081164139/)

In this case the implementation of the binary search is not so important (the application in this problem is trivial), so use the std function is more clear and could help to understand that the std function return the correct index also if the element is not in the array.

### Important

### DIFF
--- a/rust/0004-median-of-two-sorted-arrays.rs
+++ b/rust/0004-median-of-two-sorted-arrays.rs
@@ -1,56 +1,12 @@
 impl Solution {
-    pub fn find_median_sorted_arrays(mut nums1: Vec<i32>, mut nums2: Vec<i32>) -> f64 {
-        let total = nums1.len() + nums2.len();
-        let half = total / 2;
-
-        if nums1.len() > nums2.len() {
-            std::mem::swap(&mut nums1, &mut nums2);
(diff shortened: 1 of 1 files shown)


a93a:
Hi.

While I agree with you that it's not the implementation that is the focus here and it's the thinking of how to solve the problem is the focus, I think it's a good idea to keep the original solution. Since some might to follow the implemetation of the binary search as used in the video explanation.

However, I think it's a good idea to also include the version you want to commit, for people to see another way to write the code. If you can, would be great if you contributed your solution separated by comments in the same file. For example, something like:

```
// solution 1 comment
class solution {
...
...
}

// solution 2 comment 
class solution {
...
...
}
```

This way, people can follow the solution as shown in the video, and also see an example of another way of implementing the solution using builtin libraries.

Thanks.

emanuele-em:
Edited @a93a

a93a:
Awesome, thanks !

PR #3096, opened 2023-10-25
Update 0169-majority-element.cpp

MarlzRana:
The prior solution had a time and space complexity in the O(n)
The proposed solution has the same time complexity in the O(n)
The proposed solution has an improved space complexity in the O(1)

- **File(s) Modified**: _0169-majority-element.cpp_
- **Language(s) Used**: _cpp_
- **Submission URL**: _https://leetcode.com/problems/majority-element/submissions/1083722552/_

### DIFF
--- a/cpp/0169-majority-element.cpp
+++ b/cpp/0169-majority-element.cpp
@@ -7,8 +7,8 @@ The majority element is the element that appears more than ⌊n / 2⌋ times. Yo
 class Solution {
 public:
     int majorityElement(vector<int>& nums) {
-        int count{};
-        int res{};
+        int count = 0;
+        int res = 0;
 
         for (const int& num: nums) {
             if (count == 0) {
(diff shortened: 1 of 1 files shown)


cpp/0169-majority-element.cpp
```diff
@@ -7,12 +7,16 @@ The majority element is the element that appears more than ⌊n / 2⌋ times. Yo
...
     int majorityElement(vector<int>& nums) {
-        unordered_map<int, int> mp;
-        int n = nums.size();
-        for (int& i : nums){
-           if(++mp[i] > n/2)
-               return i;
-       }
-        return -1;
+        int count{};
+        int res{};
```
MHamiid, line 11: Could you please replace these with zero initializations to be more clear?

MarlzRana: This is technically zero initialization (see https://www.learncpp.com/cpp-tutorial/variable-assignment-and-initialization/).

I have replaced it with the proposed zero assignment initialization as requested (if I understand the proposed change correctly).

MHamiid: Yes, you are correct, but I believe this is more clear that way as not to get any one confused.

PR #3099, opened 2023-10-26
Update 0208-implement-trie-prefix-tree.py

cwillett:
Please be so kind as to consider my PR:
It is better to use `is None` instead of `== None` when checking for None in Python. The reason is that `is` checks for identity (whether the two objects are the same object in memory), whereas `==` checks for equality (whether the two objects have the same value).

In Python, `None` is a singleton object, and using `is` to check for `None` is more efficient and idiomatic. It ensures that you are specifically checking if a variable is bound to the `None` object and not just a value that happens to be equivalent to `None`.
Thanks!

- **File(s) Modified**: python/0208-implement-trie-prefix-tree.py
- **Language(s) Used**: python
- **Submission URL**: https://leetcode.com/problems/implement-trie-prefix-tree/submissions/1094980074/

### Important

### DIFF
--- a/python/0208-implement-trie-prefix-tree.py
+++ b/python/0208-implement-trie-prefix-tree.py
@@ -18,7 +18,7 @@ class Trie:
         curr = self.root
         for c in word:
             i = ord(c) - ord("a")
-            if curr.children[i] == None:
+            if curr.children[i] is None:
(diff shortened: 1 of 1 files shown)


Ykhan799:
@cwillett Thanks for your contribution 👍

PR #3114, opened 2023-11-05
Update 0007-reverse-integer.c with proper solution

Wirgiliusz:
- **File(s) Modified**: 0007-reverse-integer.c
- **Language(s) Used**: C
- **Submission URL**: https://leetcode.com/problems/reverse-integer/submissions/1092382115/

### Explanation:
The problem description states: 
> **Assume the environment does not allow you to store 64-bit integers (signed or unsigned).**

But the current solution clearly abuses this by using "long" type which turns out to be defined as larger than 32bit on LeetCode (probably 64bit).
This allows for a little shortcut by inversing and storing the whole number (even if larger than 32bit) and then comparing if it is larger than INT_MAX or lower than INT_MIN. But as shown above this is explicitly prohibited by the problem description, so it should not be used as accepted answer.

The solution that I have posted follows the classic pattern also explained in the video.

### DIFF
--- a/c/0007-reverse-integer.c
+++ b/c/0007-reverse-integer.c
@@ -1,12 +1,20 @@
-int reverse(int x) {
-    long reversed_num = 0;
-    while (x != 0) {
-        reversed_num = reversed_num * 10 + x % 10;
-        x /= 10;
+int reverse(int x){
+    int res = 0;
+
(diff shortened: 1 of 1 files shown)


a93a:
Good catch! Thank you. I've merged the PR now.

PR #3605, opened 2024-08-14
Cleaner Solution for problem 0153

Karthik-PM:
The code to this problem is very unclean, there are unnecessary if statement in the first section of the while loop, with a break statement. The value of res is being written more than once in every case, even though, its only is needed to be put once. All of this makes the solution less understandable and unclean.

given solution:

image

My solution:
image


Ykhan799:
@Karthik-PM Good Suggestion 👍

PR #3607, opened 2024-08-18
covers edge case for top k frequent elements solution

ahsan722505:
### Before:
inputs: [1,1,2,2,3,3] , 2
output: [ 1, 2, 3 ]

### After:
inputs: [1,1,2,2,3,3] , 2
output: [ 1, 2 ]


Ykhan799:
@ahsan722505 This change is not necessary. The test case you are suggesting is invalid when I input it into leetcode.

PR #3649, opened 2024-10-05

Srihari2222:
@neetcode-gh 
- Took 5 hours for this work.
- Please review this.
- Also, these problems doesn't have heap involved but please inform that should I add custom heap template for JS code in future problems or leave like in this article: [this one](https://github.com/neetcode-gh/leetcode/blob/main/articles/top-k-elements-in-list.md) .

### DIFF
--- a/articles/buy-and-sell-crypto.md
+++ b/articles/buy-and-sell-crypto.md
@@ -1,0 +1,295 @@
+## 1. Brute Force
+
+::tabs-start
+
+```python
+class Solution:
+    def maxProfit(self, prices: List[int]) -> int:
+        res = 0
+        for i in range(len(prices)):
+            buy = prices[i]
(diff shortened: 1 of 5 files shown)


neetcode-gh:
Thanks @Srihari2222 this looks good.

---

I currently do have a heap class and a queue class available in js. Here are the ones I'm using:

https://github.com/datastructures-js/priority-queue/
https://github.com/datastructures-js/queue/tree/master

They are the same as the ones on LC, but mine have a higher version, so they may not work on LC. Just make sure they work on NC.

I updated the top-k-elements JS heap solution here: https://github.com/neetcode-gh/leetcode/commit/ec05954c05717d454a37855b734017b3a800d239

In future problems requiring a heap, you can handle the JS solutions similarly. Let me know if you have any questions.

PR #3661, opened 2024-10-07

Srihari2222:
@neetcode-gh 
- Took  5hours 30mins for this work as there are many approaches and the number of lines went to nearly 3000.
-  Please review this.

### DIFF
--- a/articles/binary-search.md
+++ b/articles/binary-search.md
@@ -1,0 +1,496 @@
+## 1. Recursive Binary Search
+
+::tabs-start
+
+```python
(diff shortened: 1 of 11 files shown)


neetcode-gh:
Thanks, looks good. Just to confirm, you checked that all of these work on NC right? @Srihari2222

Srihari2222:
@neetcode-gh 
 Every single code that is present in these articles works perfectly fine on neetcode.io. Like, I have tested every code after before requesting a PR.

PR #3665, opened 2024-10-09
Refactor 0347-top-k-frequent-elements.go

obaibula:
- **File(s) Modified**: go/0347-top-k-frequent-elements.go
- **Language(s) Used**: Go
- **Submission URL**: https://leetcode.com/problems/top-k-frequent-elements/submissions/1416811237/

1. The named return value in this case is confusing.
2. Fixed overhead in filling the countMap.
3. The built-in len function in Go for slices calculates on every call. It is best to store the result in a variable.
4. It is sufficient to check if len(res) != k in the for-loop condition.


Ykhan799:
@obaibula Good refactoring, thanks for your contribution 👍

PR #3666, opened 2024-10-09

Srihari2222:
@neetcode-gh 

- Took 6 hours for this work.
- Please review this.

### DIFF
--- a/articles/add-two-numbers.md
+++ b/articles/add-two-numbers.md
@@ -1,0 +1,437 @@
+## 1. Recursion
+
+::tabs-start
+
+```python
+# Definition for singly-linked list.
+# class ListNode:
+#     def __init__(self, val=0, next=None):
+#         self.val = val
+#         self.next = next
+
+class Solution:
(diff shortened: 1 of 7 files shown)


neetcode-gh:
Thanks @Srihari2222 looks good, just some minor changes:

1. Can we rename the "Tortoise And Hare Algorithm" solution to "Fast and Slow Pointers"
2. For the brute force solution of remove-node-from-end-of-list, can we make the time complexity variables lower case, just for consistency. e.g. O(N) to O(n)

This is really minor, but might be worth doing.

Srihari2222:
@neetcode-gh 
yeah, I will rename it. But for the O(n), should I write the comment as the problem has n as a parameter?

Srihari2222:
@neetcode-gh 
I have renamed the name.
Waiting for the confirmation regarding the O(N).

neetcode-gh:
@Srihari2222 good point, capital N is fine in that case.

PR #3679, opened 2024-10-13
Improve 0036-valid-sudoku.go

obaibula:
- **File(s) Modified**: go/0036-valid-sudoku.go
- **Language(s) Used**: Go
- **Submission URL**: https://leetcode.com/problems/valid-sudoku/submissions/1421082466/

The previous solution did not match the video explanation.

The new solution has been adjusted to align with the video, but it uses arrays instead of maps. Here’s why:

Since Go does not have a set structure, we would typically use a map, specifically map[int]map[byte]bool. However, if the inner map is not initialized with the make function, attempting to insert values into a nil map will result in a panic. Therefore, in each iteration, we must check whether map[byte]bool is nil and create it if necessary.

This issue can be resolved by using arrays, which is why I decided to adjust the solution to use arrays instead of map[int]map[byte]bool.

The map-based solution would require three additional if statements, as shown in this submission: https://leetcode.com/problems/valid-sudoku/submissions/1421056544/.

### DIFF
--- a/go/0036-valid-sudoku.go
+++ b/go/0036-valid-sudoku.go
@@ -1,34 +1,25 @@
-// Time Complexity O(n2)
 func isValidSudoku(board [][]byte) bool {
-    
-    hashMap := make(map[string]bool)
-    
-    for i:=0; i<9; i ++ {
-        for j:=0;j<9; j++ {
-            
-            row:= i
-            column :=j
-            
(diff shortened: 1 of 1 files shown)


Ykhan799:
@obaibula Good refactoring, Thanks for your contribution 👍

PR #3712, opened 2024-11-03
Sri Hari: Batch-3/Added Golang, Kotlin

Srihari2222:
@neetcode-gh 

- Took 10 hours for this work.
- It is taking 10 to 15 minutes for each article.
- Please review it and give feedback and suggestions.
- I used leetcode for testing.
- But maintained the function name as per Neetcode only.
- There are not many data structures in Golang so I used library mentioned in leetcode, so cross check it.
- Also, while creating environment for these languages, import necessary packages.

---
Waiting for your feedback.

### DIFF
--- a/articles/add-two-numbers.md
+++ b/articles/add-two-numbers.md
@@ -226,6 +226,83 @@ public class Solution {
 }
 ```
 
+```go
+/**
+ * Definition for singly-linked list.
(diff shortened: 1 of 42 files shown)


neetcode-gh:
Looks good thanks

PR #3740, opened 2024-11-14

Srihari2222:
@neetcode-gh 

- Took 5 hours for this work. 
- Please review it.

---
- Roughly 20minutes for an article based on the problem.

---
- When we have no chance other than recommending the optimal solution, I used this line 
"You have to aim for the solution with "XYZ" complexity".

- But when I suggest sub-optimal solution, I used 
"You have to aim for the solution as good or better than "XYZ" complexity".

### DIFF
--- a/hints/anagram-groups.md
+++ b/hints/anagram-groups.md
@@ -10,7 +10,7 @@
 <details class="hint-accordion">  
     <summary>Hint 1</summary>
     <p>
-    A naive solution would be to sort each string and group them using a hash map. Can you think of a better way?
+    A naive solution would be to sort each string and group them using a hash map. This would be an <code>O(m * nlogn)</code> solution. Can you think of a better way?
     </p>
 </details>
 
--- a/hints/anagram-groups.md
+++ b/hints/anagram-groups.md
@@ -10,7 +10,7 @@
 <details class="hint-accordion">  
     <summary>Hint 1</summary>
     <p>
-    A naive solution would be to sort each string and group them using a hash map. This would be an <code>O(m * nlogn)</code> solution. Can you think of a better way?
+    A naive solution would be to sort each string and group them using a hash map. This would be an <code>O(m * nlogn)</code> solution. Though this solution is acceptable, can you think of a better way without sorting the strings?
     </p>
 </details>
 
@@ -26,6 +26,6 @@
 <details class="hint-accordion">  
     <summary>Hint 3</summary>
     <p>
-    We can simply use an <code>O(26)</code> size array to count the frequency of each character in a string. Then, we can use this array as the key in a hash map to group the strings.
+    We can simply use an array of size <code>O(26)</code>, since the character set is <code>a</code> through <code>z</code> (<code>26</code> continuous characters), to count the frequency of each character in a string. Then, we can use this array as the key in the hash map to group the strings.
     </p>
 </details>
--- a/hints/anagram-groups.md
+++ b/hints/anagram-groups.md
@@ -1,0 +1,31 @@
+<br>
+<details class="hint-accordion">  
+    <summary>Recommended Time & Space Complexity</summary>
+    <p>
+    You should aim for a solution with <code>O(m * n)</code> time and <code>O(m)</code> space, where <code>m</code> is the number of strings and <code>n</code> is the length of the longest string.
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 1</summary>
+    <p>
+    A naive solution would be to sort each string and group them using a hash map. Can you think of a better way?
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 2</summary>
+    <p>
+    By the definition of an anagram, we only care about the frequency of each character in a string. How is this helpful in solving the problem?
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 3</summary>
+    <p>
+    We can simply use an <code>O(26)</code> size array to count the frequency of each character in a string. Then, we can use this array as the key in a hash map to group the strings.
+    </p>
+</details>
--- a/hints/duplicate-integer.md
+++ b/hints/duplicate-integer.md
@@ -10,7 +10,7 @@
 <details class="hint-accordion">  
     <summary>Hint 1</summary>
     <p>
-    A brute force solution would be to check every element against every other element in the array. This would be an <code>O(n<sup>2</sup>)</code> solution. Can you think of a better way?
+    A brute force solution would be to check every element against every other element in the array. This would be an <code>O(n^2)</code> solution. Can you think of a better way?
     </p>
 </details>
 
--- a/hints/duplicate-integer.md
+++ b/hints/duplicate-integer.md
@@ -1,7 +1,8 @@
 <br>
 <details class="hint-accordion">  
     <summary>Recommended Time & Space Complexity</summary>
-    <p>You should aim for a solution as good or better than <code>O(n)</code> time and <code>O(n)</code> space, where <code>n</code> is the size of the input array.
+    <p>
+    You should aim for a solution with <code>O(n)</code> time and <code>O(n)</code> space, where <code>n</code> is the size of the input array.
     </p>
 </details>
 
@@ -9,7 +10,7 @@
 <details class="hint-accordion">  
     <summary>Hint 1</summary>
     <p>
-    A brute force solution would be to check every element against every other element in the array. This would be an <code>O(n^2)</code> solution. Can you think of a better way?
+    A brute force solution would be to check every element against every other element in the array. This would be an <code>O(n<sup>2</sup>)</code> solution. Can you think of a better way?
     </p>
 </details>
 
--- a/hints/is-palindrome.md
+++ b/hints/is-palindrome.md
@@ -10,7 +10,7 @@
 <details class="hint-accordion">  
     <summary>Hint 1</summary>
     <p>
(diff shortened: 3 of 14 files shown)


hints/duplicate-integer.md
```diff
@@ -1,15 +1,16 @@
...
+    You should aim for a solution with <code>O(n)</code> time and <code>O(n)</code> space, where <code>n</code> is the size of the input array.
     </p>
 </details>
 
 <br>
 <details class="hint-accordion">  
     <summary>Hint 1</summary>
     <p>
-    A brute force solution would be to check every element against every other element in the array. This would be an <code>O(n^2)</code> solution. Can you think of a better way?
+    A brute force solution would be to check every element against every other element in the array. This would be an <code>O(n<sup>2</sup>)</code> solution. Can you think of a better way?
```
neetcode-gh, line 13: Unfortunately, there's a limitation in the way my site renders markdown. The <sup></sup> wont work as intended so I would leave this as-is for now. And do the same for all other files.

This isn't your fault. I appreciate you trying to improve this though!

hints/is-palindrome.md
```diff
@@ -0,0 +1,31 @@
...
+    <p>
+    You should aim for a solution with <code>O(n)</code> time and <code>O(1)</code> space, where <code>n</code> is the length of the input string.
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 1</summary>
+    <p>
+    A brute force solution would be to create a copy of the string, reverse it, and then check for equality. This would be an <code>O(n)</code> solution with extra space. Can you think of a better way?
```
neetcode-gh, line 13: Minor: We can be more specific by saying "Can you think of a way to do this without O(n) space?"

"Better" is a bit ambiguous.

hints/longest-consecutive-sequence.md
```diff
@@ -0,0 +1,31 @@
...
+    <p>
+    A brute force solution would be to consider every element from the array as the start of the sequence and count the length of the sequence formed with that starting element. This would be an <code>O(n<sup>2</sup>)</code> solution. Can you think of a better way?
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 2</summary>
+    <p>
+    Is there any repeated work we are doing?
```
neetcode-gh, line 21: Can be more specific, e.g. 

Is there any way to identify the start of a sequence? For example, in [1,2,3,10,11,12], only 1 and 10 are the beginning of a sequence.

hints/longest-consecutive-sequence.md
```diff
@@ -0,0 +1,31 @@
...
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 3</summary>
+    <p>
+    Yes, we should only start building the sequence with an element <code>num</code> such that <code>num - 1</code> is not present in the array. We can use a hash set for fast lookups.
```
neetcode-gh, line 29: Reword this one, after updating above.

hints/max-water-container.md
```diff
@@ -0,0 +1,31 @@
+<br>
+<details class="hint-accordion">  
+    <summary>Recommended Time & Space Complexity</summary>
+    <p>
+    You should aim for a solution with <code>O(n<sup>2</sup>)</code> time and <code>O(1)</code> space, where <code>n</code> is the size of the input array.
```
neetcode-gh, line 5: Shouldn't this be O(n)?

hints/products-of-array-discluding-self.md
```diff
@@ -0,0 +1,39 @@
+<br>
+<details class="hint-accordion">  
+    <summary>Recommended Time & Space Complexity</summary>
+    <p>
+    You should aim for a solution with <code>O(n)</code> time and <code>O(1)</code> space, where <code>n</code> is the size of the input array.
```
neetcode-gh, line 5: For this one i think it's fine to have O(n) space since it's only a medium

hints/products-of-array-discluding-self.md
```diff
@@ -0,0 +1,39 @@
...
+    <p>
+    A brute-force solution would be to iterate through the array for each <code>nums[i]</code> and compute the product of the array except for that element. This would be an <code>O(n<sup>2</sup>)</code> solution. Can you think of a better way?
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 2</summary>
+    <p>
+    Is there a way to avoid the repeated work? Maybe there is a technique with linear time?
```
neetcode-gh, line 21: I would replace "Maybe there is a technique with linear time?" with "Maybe we can store the results of the repeated work in an array."

hints/three-integer-sum.md
```diff
@@ -0,0 +1,39 @@
...
+    <p>
+    We can fix index <code>i</code> and get <code>nums[i] = -(nums[j] + nums[k])</code> after rearranging the equation, making <code>target = -(nums[j] + nums[k])</code>. For each index <code>i</code>, we can compute the required <code>target</code>. Which algorithm is suitable for finding the <code>target</code> pairs?
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 4</summary>
+    <p>
+    To efficiently find the <code>j</code> and <code>k</code> pairs, we can use the two pointer algorithm.
```
neetcode-gh, line 37: Might be worth mentioning we run the two pointer approach on all elements to the right of i.

hints/two-integer-sum-ii.md
```diff
@@ -0,0 +1,31 @@
...
+    <p>
+    A brute force solution would be to check every pair of numbers in the array. This would be an <code>O(n<sup>2</sup>)</code> solution. Can you think of a better way?
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 2</summary>
+    <p>
+    Can you think of an algorithm by taking the advantage of array being sorted?
```
neetcode-gh, line 21: Maybe we can be more specific:

If nums[0] + nums[n-1] > target, than we know nums[n-1] can not possibly be included in any pairs. Why? Because nums[n-1] is the largest element in the array. Even by adding it with nums[0], which is the smallest element, we still exceed the target.


---

This hints at a two pointer solution, and also gives some intuition of why it works.

hints/two-integer-sum.md
```diff
@@ -0,0 +1,31 @@
...
+    <p>
+    Given, We need to find indices <code>i</code> and <code>j</code> such that <code>i != j</code> and <code>nums[i] + nums[j] == target</code>. Can you rearrange the equation and try to fix any index to iterate on? 
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 3</summary>
+    <p>
+    We can fix index <code>i</code> and iterate on it. The equation becomes <code>nums[i] == target - nums[j]</code> when we rearrange it. Let <code>value = target - nums[j]</code> and check if <code>value</code> exists in the hash map as we iterate through the array, else store the current element in the hashmap with its index and continue.
```
neetcode-gh, line 29: replace `value` with `difference`.

> We can fix index <code>i</code> and iterate on it. 

The above line seems unnecessary and a little bit meaningless.

Would it be better to remove that, and say something like: we can iterate through nums with index i. 

We should also mention that we use a hashmap for O(1) lookups. I know it's implied, but since this is a beginner problem.

neetcode-gh (changes requested):
Overall, this is a good start.

I think most of my feedback centers around:

1. Being more specific
2. Being more beginner friendly

I think we should lean towards giving more generous hints.

neetcode-gh:
Thanks @Srihari2222 

Just curious what 

> Roughly 20minutes for an article based on the problem.

is in reference to?

hints/anagram-groups.md
```diff
@@ -0,0 +1,31 @@
...
+    <p>
+    By the definition of an anagram, we only care about the frequency of each character in a string. How is this helpful in solving the problem?
+    </p>
+</details>
+
+<br>
+<details class="hint-accordion">  
+    <summary>Hint 3</summary>
+    <p>
+    We can simply use an <code>O(26)</code> size array to count the frequency of each character in a string. Then, we can use this array as the key in a hash map to group the strings.
```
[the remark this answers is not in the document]
Srihari2222: Yeah, I will be more beginner-friendly and specific towards the details.

hints/duplicate-integer.md
Srihari2222: Ok, I will keep that as it is in all the files.

hints/is-palindrome.md
Srihari2222: A great point to mention for a beginner user. I will be more specific.

hints/products-of-array-discluding-self.md
Srihari2222: I thought it would be better to have a solution from your video, but fine, now I understood what you are trying to say. I will be more beginner friendly.

hints/three-integer-sum.md
Srihari2222: Yes, I  will do that. But I have added 5 hints only for this problem as we don't want to ask a user a question and give answer in the same hint.

Srihari2222:
> Thanks @Srihari2222
> 
> Just curious what
> 
> > Roughly 20minutes for an article based on the problem.
> 
> is in reference to?

@neetcode-gh 
I was also checking for grammatical errors in the hints. But now I can see that I should be more obvious  and simple. I meant 20min as hard problems can take. It was the max limit but, I will make sure and try hard to wrap it up in 10 to 15mins at max.
As, your feedback made me more clear about things, It will be easy for me to make hints for further problems.

neetcode-gh:
@Srihari2222 Thanks and great work!

PR #3742, opened 2024-11-17

Srihari2222:
@neetcode-gh 
- Took 5 hours for this work.
- Please review it.

---
For classic problems like "Linked list cycle detection"  which approach should I recommend? Should I recommend an approach that is meant for the problem or should I go with the one that is basic and passes? Suggest between "Hash set" versus "fast and slow pointers" solutions.

Thank you.


neetcode-gh:
@Srihari2222 For the cycle detection I think its fine to recommend two pointer solution.

For Find the Duplicate number problem, I think its fine to let them use extra space or let them modify the input array.

PR #3746, opened 2024-11-22

Srihari2222:
@neetcode-gh 
- Took 5 hours for this work.
- Please review it.

---

Can you suggest the complexity for the problem "Design twitter", as it comes under heap category but the sorting solution also passes greatly for this problem?

### DIFF
--- a/articles/task-scheduling.md
+++ b/articles/task-scheduling.md
@@ -341,7 +341,7 @@ class Solution {
 
 ---
 
-## 2. Heap
+## 2. Max-Heap
 
 ::tabs-start
 
@@ -609,7 +609,7 @@ class Solution {
 ### Time & Space Complexity
 
 * Time complexity: $O(m)$
-* Space complexity: $O(m)$
+* Space complexity: $O(1)$
 
 > Where $m$ is the number of tasks.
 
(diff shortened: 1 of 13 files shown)


articles/task-scheduling.md
```diff
@@ -609,7 +609,7 @@ class Solution {
 ### Time & Space Complexity
 
 * Time complexity: $O(m)$
-* Space complexity: $O(m)$
+* Space complexity: $O(1)$
```
neetcode-gh, line 612: Maybe i'm mistaken, but shouldn't the space complexity remain O(m)?

Oh, because the character set is 26. Can we add a note that this is because we only store a limited amount of values in the heap, like in: https://github.com/neetcode-gh/leetcode/commit/7a6051c623e905a01768115b1b8a5ba3963d1c32

Someone asked me a question about this the other day so i thought it was worth clarifying.

neetcode-gh (changes requested):
Looks good, minor change requested

articles/task-scheduling.md
Srihari2222: The space complexity is O(1) because we store almost 26 elements only in any data structure.

It is a good idea to add that note. I will do that for all articles when I am adding the "Swift" code.

PR #3749, opened 2024-11-24

Srihari2222:
@neetcode-gh 
- Took 5 hours for this work
- Please review it.

---
Can you suggest the complexity for the problem "Design twitter", as it comes under heap category but the sorting solution also passes greatly for this problem?


neetcode-gh:
@Srihari2222 We can recommend the heap solution I think, since it's medium sorting might be too easy

PR #3751, opened 2024-11-26
Sri Hari: Batch-3/Neetcode-ALL/Added articles

Srihari2222:
@neetcode-gh 

- Took 10 hours for this work.
- Please review it.

---
- I will be doing the hints and articles tasks parallelly.
- I added class-based solution for Javascript.
- I noticed you have created new URL for  the solutions. That looks good but the loading time is somewhat long compared to other platforms such as gfg and takeUforward.

### DIFF
--- a/articles/longest-common-prefix.md
+++ b/articles/longest-common-prefix.md
@@ -1,0 +1,503 @@
+## 1. Iteration - I
+
+::tabs-start
+
+```python
+class Solution:
+    def longestCommonPrefix(self, strs: List[str]) -> str:
+        prefix = strs[0]
+        for i in range(1, len(strs)):
+            j = 0
+            while j < min(len(prefix), len(strs[i])):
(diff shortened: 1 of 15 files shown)


articles/longest-common-prefix.md
```diff
@@ -0,0 +1,503 @@
+## 1. Iteration - I
```
neetcode-gh, line 1: minor: for some of the solution titles i think we can be more descriptive. For example the LC editorial mentions horizontal scanning vs vertical scanning: https://leetcode.com/problems/longest-common-prefix/editorial/

neetcode-gh (approved):
Looks good thanks

articles/longest-common-prefix.md
Srihari2222: I will modify that in next PR.

PR #3761, opened 2024-12-06
Clarify Space Complexity For DFS Traversal Of Binary Trees

reywilliams:

- **File(s) Modified**: 
  - `articles/balanced-binary-tree.md`
  - `articles/binary-tree-diameter.md`
  - `articles/depth-of-binary-tree.md`
  - `articles/same-binary-tree.md`
- **Language(s) Used**: `md`
- **Submission URL**: N/A

> [!NOTE]  
> All references to $\log$ here are $\log_2$ (as we usually assume with CS and its binary nature lol 😅 )

Sorry if this is a D1 yap sesh - I thought these would be helpful clarifications and learning and sharing them help me understand. I can add these to other DFS approaches as well, but I honestly didn't feel comfortable as I didn't use them or look over them. Should still be applicable as it is DFS though.

I wanted to clarify some space complexities for the DFS traversal of binary trees. This might help in evaluating what approach to use during an interview or the complexities of your chosen method.

From **my** understanding (please challenge this if need be), the DFS traversal of binary trees has space complexities as follows:

* Space complexity: $O(h)$  
  * Best Case ([balanced tree](https://www.geeksforgeeks.org/balanced-binary-tree/)): $O(log(n))$
  * Worst Case ([degenerate tree](https://www.geeksforgeeks.org/introduction-to-degenerate-binary-tree/)): $O(n)$

> Where $n$ is the number of nodes in the tree and $h$ is the height of the tree.

### Best Case (Balanced Tree)

See this balanced binary tree here. 
```
        1 - level 🥇 
       / \ 
      2   3 - level 🥈 
     / \ / \
    4  5 6  7 - level 🥉 
```

In this case, the height of the tree ( $O(h)$ ) can be represented by $O(log(n))$. This tree's height is $3$ and we know that $O(log_2(7))$ is around $3$. We can make this "representation" as every node (except the 🍃 ) usually has two children.
With this height, the largest the call stack can go to is $h$ or $3$

### Worst Case (Degenerate Tree / Linear Tree)

See this degenerate binary tree here (he's not a bad guy, just unbalanced 🥁 )
```  
        1
         \
          2
           \
            3
             \
              4
               \
                5 
```

In this case, the height of the tree ( $O(h)$ ) can be represented by $O(n)$ as every parent has one child. With this height, the largest the call stack can go to is $n$ or $5$.

### DIFF
--- a/articles/balanced-binary-tree.md
+++ b/articles/balanced-binary-tree.md
@@ -526,7 +526,11 @@ class Solution {
 ### Time & Space Complexity
 
 * Time complexity: $O(n)$
-* Space complexity: $O(n)$
+* Space complexity: $O(h)$  
+  * Best Case ([balanced tree](https://www.geeksforgeeks.org/balanced-binary-tree/)): $O(log(n))$
+  * Worst Case ([degenerate tree](https://www.geeksforgeeks.org/introduction-to-degenerate-binary-tree/)): $O(n)$
+
+> Where $n$ is the number of nodes in the tree and $h$ is the height of the tree.
 
 ---
 
--- a/articles/binary-tree-diameter.md
+++ b/articles/binary-tree-diameter.md
@@ -520,7 +520,11 @@ class Solution {
 ### Time & Space Complexity
 
 * Time complexity: $O(n)$
-* Space complexity: $O(n)$
+* Space complexity: $O(h)$  
+  * Best Case ([balanced tree](https://www.geeksforgeeks.org/balanced-binary-tree/)): $O(log(n))$
+  * Worst Case ([degenerate tree](https://www.geeksforgeeks.org/introduction-to-degenerate-binary-tree/)): $O(n)$
+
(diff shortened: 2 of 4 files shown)


Srihari2222:
@reywilliams 

Thanks for your research. The one thing that is challenging is to assign complexities that vary based on the intuition. For temporarily I have added the worst-case complexities and I'm also not 100% sure that all complexities are picture perfect (but doesn't lead to a wrong direction though). 
We can't add links from other websites like GFG. Let's wait for Navi's response on this. 

---
Once again, thank you for the work, there are slight modifications needed for complexities for Neetcode-150. I am adding articles for all the problems and will also improve them in iterative manner.
I appreciate your feedback.

reywilliams:
That's very fair, thanks for taking a look @Srihari2222! Looking forward to any follow-up.

neetcode-gh:
@reywilliams Thanks, yeah this is a useful clarification I think.

PR #3767, opened 2024-12-17
Sri Hari: Batch-4/Neetcode-All/Added-articles

Srihari2222:
@neetcode-gh 

- Took 25 hours for this work.
- Please review it.

---
For the problem, LC: 2707 Extra Characters in a string, I have added more than 6 approaches. Can I add more 3 approaches which include Rolling hash, and a very unique solution with O(nlogn) time (DP + Rolling Hash + Segment Tree) which is nowhere in the discussion section?

### DIFF
--- a/articles/buy-two-chocolates.md
+++ b/articles/buy-two-chocolates.md
@@ -1,0 +1,211 @@
+## 1. Brute Force
+
+::tabs-start
+
+```python
+class Solution:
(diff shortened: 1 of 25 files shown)


neetcode-gh:
@Srihari2222 Thanks Sri! I think while the extra solutions would def be useful to some people, but from the metrics I have I think the bigger priority at the moment would just be to have an article for each problem.

After that we can always make improvements.

PR #3972, opened 2025-03-29
Create 1269-number-of-ways-to-stay-in-the-same-place-after-some-steps.js

aadil42:
Solved number-of-ways-to-stay-in-the-same-place-after-some-steps

File(s) Added: 1269-number-of-ways-to-stay-in-the-same-place-after-some-steps.js
Language(s) Used: JavaScript
Submission URL : https://leetcode.com/problems/number-of-ways-to-stay-in-the-same-place-after-some-steps/submissions/1589993538/

### DIFF
--- a/javascript/1269-number-of-ways-to-stay-in-the-same-place-after-some-steps.js
+++ b/javascript/1269-number-of-ways-to-stay-in-the-same-place-after-some-steps.js
@@ -1,0 +1,29 @@
+/**
+ * DP | Recursion
+ * Time O(n^2) | Space O(n^2)
+ * https://leetcode.com/problems/number-of-ways-to-stay-in-the-same-place-after-some-steps/
+ * @param {number} steps
+ * @param {number} arrLen
+ * @return {number}
+ */
+var numWays = function(steps, arrLen) {
+    
+    const cache = new Map();
+    const mod = 10**9 + 7;
+
+    const dfs = (i, leftSteps) => {
+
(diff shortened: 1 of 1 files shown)


javascript/1269-number-of-ways-to-stay-in-the-same-place-after-some-steps.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * DP | Recursion
+ * Time O(n^2) | Space O(n^2)
```
aakhtar3, line 3: Can reduce space

```js
/**
 * DP - Bottom Up
 * Array - Tabulation
 * Time O(N * MIN(N, M)) | Space O(MIN(N, M))
 * @param {number} steps
 * @param {number} arrLen
 * @return {number}
 */
var numWays = (steps, arrLen) => {
    const mod = ((10 ** 9) + 7);
    const length = Math.min(((steps >> 1) + 1), arrLen);

    let tabu = initTabu(length);
    tabu[1] = 1;

    for (let step = steps; (0 < step); step--) {
        const prev = initTabu(length);

        for (let pos = 1; (pos <= length); pos++) {
            prev[pos] = (
                tabu[pos - 1]
                + tabu[pos]
                + tabu[pos + 1]
            ) % mod;
        }

        tabu = prev;
    }

    return tabu[1];    
};

var initTabu = (length) => new Array(length + 2).fill(0);
```

PR #5494, opened 2026-02-19
solution of 1888 in java

adityasam612:


adityasam612:
adding up the solution of java of 1888

PR #6033, opened 2026-08-01
Fix 62 article solution snippets that fail to compile on the NeetCode judge

neetcode-gh:
## Summary

62 solution snippets across 33 articles used libraries or type signatures that don't exist in the NeetCode judge environment — every one of them failed to compile for any user who copied it into the editor. All are now fixed and **verified green against the production judge** via the article-solutions harness in the main repo.

Three bug classes:

1. **swift-collections types with no import and no judge support** (53 snippets): `Heap<T>` (22) and `Deque<T>` (31). Heaps now use the comparator-based `struct Heap<T>` that this repo's own Swift reference solutions already define (the house idiom, already inlined by several articles); Deques get a head-index `struct Deque<T>` shim with amortized O(1) `popFirst` — chosen over plain arrays because these are BFS queues and `Array.removeFirst()` is O(n). Zero call-site changes to published code.
2. **gods library imports in Go** (2): trapping-rain-water's Stack tab (`linkedliststack`) → plain slice; meeting-schedule-ii's Min Heap (`priorityqueue`) → stdlib `container/heap`.
3. **LeetCode-style signatures vs the NeetCode drivers** (7): meeting-schedule-ii (4 Rust tabs, 2 Kotlin tabs) and meeting-schedule (2 Rust tabs) now use the drivers' `Interval` type; plus clone-graph's pre-existing `[Node: Node]` dictionary on a non-Hashable class (both Swift tabs) → `ObjectIdentifier` keys.

Left alone deliberately: 6 `Heap` snippets and 3 gods usages in articles whose problems have no judge directory (list in the commit messages), and all intended-tier Brute Force TLEs.

## Verification
Every fixed snippet ran against the production judge (`test-problem-article-solutions.ts --article-root <this branch>`): 62/62 pass, and every untouched approach in the affected articles stayed green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/articles/binary-tree-right-side-view.md
+++ b/articles/binary-tree-right-side-view.md
@@ -686,6 +686,45 @@ class Solution {
         return res
     }
 }
+
+struct Deque<T>: ExpressibleByArrayLiteral {
+    private var elements: [T] = []
+    private var head = 0
+
+    init() {}
+
+    init(_ elements: [T]) {
+        self.elements = elements
+    }
+
+    init(arrayLiteral elements: T...) {
+        self.elements = elements
+    }
+
+    var isEmpty: Bool { head >= elements.count }
+    var count: Int { elements.count - head }
+
+    mutating func append(_ value: T) {
+        elements.append(value)
+    }
+
+    mutating func popFirst() -> T? {
+        guard head < elements.count else { return nil }
+        let value = elements[head]
+        head += 1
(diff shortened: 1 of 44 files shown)


neetcode-gh:
Closing — the right fix is adding swift-collections to the NC-Judge toolchain so these snippets compile as written, rather than inlining data structures per snippet.

